PR #22, opened 2022-10-20
Add dprint, dependabot and enable CI

klochowicz:


.github/workflows/add-new-issues-to-project-board.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  add_issue_to_project:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: itchysats/actions/add-project-item@main
+        with:
+          board_id: ${{ secrets.PROJECT_BOARD_ID }}
+          item_id: ${{ github.event.issue.node_id }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.ITCHY_PROJECT_AUTOMATION_TOKEN }}
```
luckysori, line 17: ❓ Do we need to add these secrets to this repository?

luckysori (approved):

.github/workflows/add-new-issues-to-project-board.yml
klochowicz: no need, these 2 are defined on org level and shared across all our repos. 
Good question tho! 👍

PR #26, opened 2022-10-20
chore: Add a simple Makefile to encapsulate formatting

klochowicz:
Unfortunately, I couldn't get dprint working for flutter.
As a stopgap, here's a simple Makefile that can be used to call both formatters
at once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+# format all files in the project
+format: dprint flutter-format
+
+dprint:
+	dprint fmt
+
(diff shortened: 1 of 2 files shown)


luckysori (approved):
Good idea! Can we add a check to CI or does it already exist?

klochowicz:
> Good idea! Can we add a check to CI or does it already exist?

I will add a dart formatter step separately from dprint today (also, need to add building both jobs to CI too).

PR #27, opened 2022-10-21
Provide instructions for building on all targets, including mobile

klochowicz:
Expand Makefile to encapsulate all the convoluted commands needed.
Update README.md to mention the Makefile


klochowicz:
merging this after last conversation with @luckysori and @da-kami to just go ahead with project setup if I am blocked (we can always tune things later on).

PR #29, opened 2022-10-21
CI improvements

klochowicz:
Add bors and 'lint-commit' job


klochowicz:
merging this after last conversation with @luckysori and @da-kami to just go ahead with project setup if I am blocked (we can always tune things later on).

PR #41, opened 2022-10-24
chore(ci): Build and test the whole app on Android

klochowicz:
So far the test doesn't do anything useful besides what was already in the
example found in rust-flutter-bridge repo; it serves as a basis to see that we
end-to-end test the application on Android.


klochowicz:
going ahead and merging this, as this is mostly taken from flutter-rust-bridge project and I would like to ensure deps can be updated to clear out the PR list.

PR #42, opened 2022-10-24
Wallet comparison

da-kami:
This is an attempt to get a deeper understanding of non-/semi-custodial wallet setups out there and a feature comparison of the wallets implementing them. 

This will help to position 10101 in the market, define our preferred setup and compare our solution to existing solutions out there.

### DIFF
--- a/docs/research/wallet-comparison.md
+++ b/docs/research/wallet-comparison.md
@@ -1,0 +1,167 @@
+## Existing Wallets
+
+List of Lightning wallet to be analysed:
+
+- Wallets to connect to your existing node
+  - [BlueWallet](https://bluewallet.io/)
+    - Needs LNDHub (hosted, or self-hosted)
+  - [Zeus](https://zeusln.app/)
+    - Can connect to LND or C-Lightning
+  - [FullyNoded](https://fullynoded.app/)
+    - Can connect to LND or C-Lightning
+- Full nodes
+  - [Breez](https://breez.technology/)
+    - Uses a forked LND version
+  - [Phoenix](https://github.com/ACINQ/phoenix)
+    - Uses a custom lighting implementation
+
+Below we have a look at these wallets and their feature set.
+If a wallet is not mentioned below it has not been analysed (yet).
+
+### Phoenix Wallet
+
+In a nutshell: Non-custodial setup and a great UX that comes with some (minimal) trust for certain features.
+
+Setup:
+
+- Non-custodial setup
+  - Run own LND node on Phone ([Eclair](https://github.com/ACINQ/eclair), ACINQ in-house lightning implementation)
+  - Channel management through ACINQ nodes
+  - Can configure own Electrum Server
+  - Cannot configure own Lightning node (outside the phone setup)
+- Where is trust involved?
+  - Initial channel opening
+  - Onboarding with BTC
+
+Pro:
+
+- Good looking and efficient UI and UX
+- One of the best onboarding UX out there
+  - One-click wallet setup
+  - Onboarding using BTC
+- BIP39 (+BIP84) 12-word wallet backup
+
+Pro/Con:
+
+- Channels management by ACINQ
+  - Slightly higher costs for more convenience
+- Channel backup managed by ACINQ
+  - "Minimal" trust for more convenience
+
+Con:
+
+- Hard to understand the underlying constraints for the Lightning setup
+- Costs for transactions are not transparent (retrospectively)
+  - Example
+    - Initially funded the wallet via BTC, had BTC worth `50.3` USD incoming.
+    - Once they were actually confirmed it changed to `49.95` USD because the fees were deducted.
+    - These fees are never visible again, when going to the details I cannot see how much fee was actually deducted and what for.
+    - A better cost breakdown would be nice to have.
+
+For details check these links:
+
+- https://medium.com/@ACINQ/introducing-phoenix-5c5cc76c7f9e
+- https://phoenix.acinq.co/faq
+
+### Blue Wallet
+
+In a nutshell: Semi- or Non-custodial (when using own LND) setup and a good UX that includes multiple wallet, but overall not as good as Phoenix.
+
+Setup:
+
+- Plug in own LND node, or use [LndHub](https://github.com/BlueWallet/LndHub)
+  - LndHub is a semi-custodial setup ("minimum trust for endusers")
+- Where is trust involved?
+  - LndHub
+
+Pro:
+
+- Multiple wallet support
+- Vaults (Not tried out, feels like a different feature unrelated to Lightning Wallet)
+- LApp Marketplace
+  - [Lapps standard](https://www.lapps.co/wallets) marketplace that is a great way to get an overview of what is out there to use your sats
+  - Does not integrate as deep as Breez, e.g. for LN Markets one can see the app within the app, but one has to register/login separately and cannot just start trading.
+  - Similar to Breez it is unclear what happens when clicking on one of the app icons in the marketplace.
+- One-click wallet create for setup, but onboarding not as straight-forward as Phoenix
+- Cross-platform: you can use BlueWallet on Desktop and Mobile.
+
+Con:
+
+- Cannot "empty the wallet", always have to keep at least some sats in for "paying fees"
+  - This is super annoying, it is literally impossible to empty this wallet.
+- Multi-wallet management
+  - The wallets are completely independent; sending money from one to the other is quite cumbersome.
+  - I would have expected that I can easily move funds from one wallet to another, but this is quite cumbersome.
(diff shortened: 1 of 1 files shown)


docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,171 @@
...
+
+Setup:
+
+- Plug in own LND node, or use [LndHub](https://github.com/BlueWallet/LndHub)
+  - LndHub is a semi-custodial setup ("minimum trust for endusers")
+- Where is trust involved?
+  - LndHub
+
+Pro:
+
```
klochowicz, line 78: IMHO a big pro: LApps marketplace. many, many features enabled thanks to that
```suggestion
- LApps marketplace
```

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,171 @@
...
+- Cannot "empty the wallet", always have to keep at least some sats in for "paying fees"
+  - This is super annoying, it is literally impossible to empty this wallet.
+- Multi-wallet management
+  - The wallets are completely independent; sending money from one to the other is quite cumbersome.
+  - I would have expected that I can easily move funds from one wallet to another, but this is quite cumbersome.
+  - Not sure what the use cases behind multiple wallets is; it does not feel like you average "bank account" app where you divide your money into different accounts for different purpose.
+- Send / Receive are not as intuitive as for Phoenix
+  - When using Phonex it becomes clear how Lightning works, even when one is new to send/receive and creating/paying invoices.
+  - The Blue wallet interface can be confusing, because it is not clear what one has to do to send/receive.
+- When using LndHub the backup is an lndhub link or a QR code. There is not 12-word seed available that can just be copied.
```
klochowicz, line 97: not sure whether this is a con, it might a security feature. clipboards (especially on Android) are not too safe to use, because other apps might reach into that without your knowledge ;/ (on iOS you get notified if an app wants to use clipboard).

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,171 @@
...
+  - How would I scan a QR code for backup with only one phone?
```
klochowicz, line 98: webcam? iPad? another backup phone?

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Fund wallet with BTC
+- Fund wallet with Fiat: Buy BTC feature enabled through MoonPay integration
+- Channel management
+  - Default
+  - Possibility to interact with Lightning node through basic interface for devs in Preferences
+
+Pro / Con:
+
+- "Marketplace" wallet integrations
```
klochowicz, line 133: IMHO that's definitely a pro - you can do more things with the wallet.

klochowicz (approved):
cool, thanks for the write-up! Added a couple of comments.

docs/research/wallet-comparison.md
da-kami: I don't want to use the phone clipboard - I want to be able to write it on a piece of paper without using any other tech ;)

docs/research/wallet-comparison.md
da-kami: I personally don't find that super convenient, but I'm happy to be convinced otherwise. Generally find it interesting how the three that I analysed so far handle backup so differently - and am wondering what path we want to follow.

docs/research/wallet-comparison.md
da-kami: I agree, Somehow :D 

The integrations are a great idea, but sometimes don't work very well - how would you complain about that with Breez? This is cool on paper, but when using it in reality it is not the best UX in the world. I also had trouble understanding what some of the services are - there is no explanation, but the app just opens when clicking on it and connects to your wallet. Not sure I want that - not sure others want that. 

Again, I don't want to talk this down, I think these are powerful and cool features, but I also think we can learn from what be made better.

docs/research/wallet-comparison.md
da-kami: OMG that feature is well hidden (It's a button `Marketplace` in the wallet view) 😅

I'll add that, just did not notice it before!

docs/research/wallet-comparison.md
klochowicz: for me this is the "wrong" section - IMHO the ability to have 3rd party integrations using an open standard is a pro (not in "pro/con" section).

as for the explanation, this hinges on the fact that the user knows what the want to use, e.g. by researching earlier the apps they would like to try:
https://www.lapps.co - here you have explanations of all these apps with comments and upvotes.

All I'm saying is that it's an established standard and by deliberately not taking this into account we might be alienating a huge potential userbase for no reason (instead of actually letting the user to decide whether they would like to engage with that or not).

docs/research/wallet-comparison.md
klochowicz: I agree that it might not be super convenient for some; probably it depends what's the use-case. For a small sums, I would actually use the QR code to just have the same wallet on my backup phone.

docs/research/wallet-comparison.md
klochowicz: case it point: there's 9 Android wallets according to https://www.lapps.co/wallets https://www.lapps.co/wallets

I see some merit in deliberately curating the apps for a polished experience, but this sounds a lot like Apple approach - and not sure whether the majority of users would actually want that. Perhaps some user interviews would be helpful.

da-kami: Fair enough, thanks for the link to the standard, I'll have another look at that and mention it in the Pros!

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,177 @@
...
+- Multiple wallet support
+- Vaults (Not tried out, feels like a different feature unrelated to Lightning Wallet)
+
+Pro / Con:
+
+- LApp Marketplace
+  - The marketplace is a great way to get an overview of what is out there to use your sats
+  - Does not integrate as deep as Breez, e.g. for LN Markets one can see the app within the app, but one has to register/login separately and cannot just start trading.
+  - Similar to Breez it is unclear what happens when clicking on one of the app icons in the marketplace.
+- One-click wallet create for setup, but onboarding not as clear and rich as Phoenix
```
holzeis, line 88: I also like these features

- Plausible deniability: This is a very nice security feature, protecting your money - even if you are physically forced.
- Watch only: I like that as well for security reasons, as I can keep my private keys on my hardware wallet. And I like to check how my BTC is doing ;)

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,177 @@
...
+  - Default
+  - Possibility to interact with Lightning node through basic interface for devs in Preferences
+
+Pro / Con:
+
+- Marketplace integrations
+  - e.g.: LN Markets: Can directly use funds to trade there
+  - App integrations are not native to the app, some work nice, some work less nice. The overall UX is not very good because it's mostly an integration of web-apps into breez.
+  - It would help to know what happens when clicking on the app icon
+    - This is more of an app-store experience similar to e.g. Umbrel; it might be better to have an overivew of what an app is and what happens when opening (installing?) it.
```
holzeis, line 141:
```suggestion
    - This is more of an app-store experience similar to e.g. Umbrel; it might be better to have an overview of what an app is and what happens when opening (installing?) it.
```

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,177 @@
...
+- Back up google/apple account
+  - Backup of the complete channel state + wallet
```
holzeis, line 143: ❓ Is this backup done after each transaction or only manually?

holzeis (approved):
Thanks for the write up 👍

docs/research/wallet-comparison.md
```diff
@@ -0,0 +1,177 @@
+## Existing Wallets
+
+List of Lightning wallet to be analysed:
+
+- Wallets to connect to your existing node
+  - [BlueWallet](https://bluewallet.io/)
+    - Needs LNDHub
```
bonomat, line 7:
```suggestion
    - Needs LNDHub (hosted, or self-hosted)
```

docs/research/wallet-comparison.md
klochowicz: well, it's inside a wallet, because you use that wallet funds with the marketplace. FWIW, in my case the button being right there in the middle of the screen was pretty easy to find ;D

docs/research/wallet-comparison.md
bonomat, line 88:
```suggestion
- One-click wallet create for setup, but onboarding not as clear and rich as Phoenix
- Cross-platform: you can use BlueWallet on Desktop and Mobile.
```

da-kami:
bors ping

PR #62, opened 2022-10-26
chore(ci): Use linux instead of macos for Android integration test job

klochowicz:
Ubuntu containers are less likely to exhaust our concurrency quotas.


klochowicz:
bors cancel

PR #70, opened 2022-10-27
Add BDK example

klochowicz:
Dart VM is calling into sync Rust without blocking the Flutter thread.

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -1,18 +1,27 @@
-use std::thread::sleep;
-use std::time::Duration;
-
-use crate::wallet;
 use crate::wallet::Wallet;
 use anyhow::anyhow;
+use anyhow::Context;
 use anyhow::Result;
+use bdk::Balance;
 use flutter_rust_bridge::ZeroCopyBuffer;
+use std::sync::Mutex;
 
-pub fn init_wallet() -> Result<Wallet> {
-    Box::new(crate::wallet::init_wallet())
+use state::Storage;
+
+static WALLET: Storage<Mutex<Wallet>> = Storage::new();
+
+pub fn init_wallet() -> Result<()> {
+    WALLET.set(Mutex::new(Wallet::new()?));
+    Ok(())
 }
 
-pub fn get_balance(wallet: Box<Wallet>) -> Result<Balance> {
-    wallet.sync()
+pub fn get_balance() -> Result<Balance> {
+    WALLET
+        .try_get()
+        .context("Wallet uninitialised")?
+        .lock()
+        .map_err(|_| anyhow!("cannot acquire wallet lock"))?
+        .sync()
 }
 
 // TODO: Remove the examples below when we're comfortable using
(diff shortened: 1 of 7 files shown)


klochowicz:
@holzeis I've rebased it mostly to see how much cleaner it would look after removing the generated code from git, feel free to merge it or just use it as a part of your work going forward; there's not too much raw value in this one :)

holzeis:
@klochowicz much better 👍

rust/src/api.rs
```diff
@@ -1,8 +1,18 @@
...
+use std::time::Duration;
+
+use crate::wallet;
 use anyhow::anyhow;
 use anyhow::Result;
-
 use flutter_rust_bridge::ZeroCopyBuffer;
 
+pub fn init_wallet() -> Result<String> {
+    println!("going to sleep");
```
bonomat, line 10: Do we have a logger already?

bonomat: I guess this ticket will cover it: https://github.com/itchysats/10101/issues/83

klochowicz: no logger yet! we will disable `println` etc via clippy lint soon when we will add one.

da-kami (commented):
One sentence with a bit more context would have helped.

My understanding: 
This PR is to test that we can call `sync` on the wallet and we see an updated balance (?)

Are you planning to merge this, or is this meant to be as a Draft PR that showcases how this will work? :)
If this is just a showcase maybe we can mark it as `Draft` PR until it is something that is mergable.

da-kami:
Earlier we talked about this work, and you mentioned that you want to drive this further by returning the wallet from the `init` function and then the wallet's lifecycle would be in control of dart and we would just call functions on the `wallet` object (struct) from dart. 

Are you planning to add this to this PR?

klochowicz:
> Earlier we talked about this work, and you mentioned that you want to drive this further by returning the wallet from the `init` function and then the wallet's lifecycle would be in control of dart and we would just call functions on the `wallet` object (struct) from dart.



> One sentence with a bit more context would have helped.
> 
> My understanding: This PR is to test that we can call `sync` on the wallet and we see an updated balance (?)
> 
> Are you planning to merge this, or is this meant to be as a Draft PR that showcases how this will work? :) If this is just a showcase maybe we can mark it as `Draft` PR until it is something that is mergable.

I didn't set it out as draft, as I was planning on merging this, so @holzeis could  build on top of it last Thursday. 
If you don't think it's mergeable, I'm happy to close it.

klochowicz:
> Earlier we talked about this work, and you mentioned that you want to drive this further by returning the wallet from the `init` function and then the wallet's lifecycle would be in control of dart and we would just call functions on the `wallet` object (struct) from dart.
> 
> Are you planning to add this to this PR?

I planned to add state management as a separate PR, because otherwise I'm needlessly blocking others. Can rework this one if you prefer this way.

klochowicz:
> One sentence with a bit more context would have helped.

thanks for the feedback! will try to be more explicit next time :)

klochowicz:
this PR has served its purpose, and there's not much usefulness in it anymore.

PR #71, opened 2022-10-27
chore: change target for generated sources

holzeis:
This PR changes the target folder of the generated source to generated_bridge. Which should make it easier to identify which sources are relevant and which can be skipped in case of a review.


klochowicz:
> This PR changes the target folder of the generated source to generated_bridge. Which should make it easier to identify which sources are relevant and which can be skipped in case of a review.

I must apologise for this one - I missed this part as it was already merged and this part was missing from the git commit description.  this is a fair explanation, please submit the PR again!!!

PR #72, opened 2022-10-27
chore: generate bridge code on cargo build

holzeis:
Integrates the flutter_rust_bridge_generator into the build.rs.

Eliminates the necessity of running make gen everything you change rust code. code generation is executed with cargo build.

https://cjycode.com/flutter_rust_bridge/feature/build_rs.html

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -1,0 +1,36 @@
+use lib_flutter_rust_bridge_codegen::config_parse;
+use lib_flutter_rust_bridge_codegen::frb_codegen;
+use lib_flutter_rust_bridge_codegen::get_symbols_if_no_duplicates;
+use lib_flutter_rust_bridge_codegen::RawOpts;
+
+const RUST_INPUT: &str = "src/api.rs";
+const RUST_OUTPUT: &str = "src/bridge_generated/bridge_generated.rs";
+const DART_OUTPUT: &str = "../lib/generated/bridge_generated.dart";
+const IOS_C_OUTPUT: &str = "../ios/Runner/bridge_generated.h";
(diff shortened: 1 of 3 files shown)


klochowicz (commented):
Interesting stuff! this has potential of making working across FFI boundary  a bit more convenient.

🤔However, I have a few open questions:
- how to deal with cache invalidation issues caused by `build.rs`? we have run into that in the past with other crate, and this slowed down CI builds 
- would this trigger multiple runs of the codegen when we're building for multiple targets at once? (like with `cargo-ldk` or `cargo-lipo`?).

klochowicz:
also, it seems to fail on the actual Android integration test (and clippy too)

rust/build.rs
```diff
@@ -0,0 +1,36 @@
...
+        // for other options use defaults
+        ..Default::default()
+    };
+    // get opts from raw opts
+    let configs = config_parse(raw_opts);
+
+    // generation of rust api for ffi
+    let all_symbols = get_symbols_if_no_duplicates(&configs).unwrap();
+    for config in configs.iter() {
+        frb_codegen(config, &all_symbols).unwrap();
```
klochowicz, line 34: this seems to pull Flutter in after all 🙁
https://github.com/itchysats/10101/actions/runs/3337012818/jobs/5522841154

klochowicz (changes requested):
after looking at CI action results, I don't think this is feasible - making our Rust builds depend on Flutter toolkit does not sound like a good idea.

holzeis:
closing that PR as running the generation in the build.rs does not seem to properly work.

PR #74, opened 2022-10-27
feat: example stream implementation

holzeis:
@klochowicz Streams are a pretty cool feature allowing us to push events to flutter. e.g. if a lightning event or a wallet sync events happens..

https://cjycode.com/flutter_rust_bridge/feature/stream.html

This PR is not intended to get merge, but rather to illustrate how we could use streams to callback to flutter.

PR #75, opened 2022-10-27
feat: bip39 wallet with stream api

holzeis:
- a wallet with a seed derived from BIP39.
- syncs wallet every minute and updates the ui over a stream.
- rudimentary wallet view showing address, balance and BIP39 seed phrase.

Based on `flutter_rust_bridge::StreamSink`

see also
https://github.com/itchysats/10101/pull/74#issue-1425681629

### DIFF
--- a/lib/wallet.dart
+++ b/lib/wallet.dart
@@ -1,0 +1,47 @@
+/// Code in this file is more about Flutter than our package, `flutter rust bridge`.
+/// To understand this package, you do not need to have a deep understanding of this file and Flutter.
+/// Thus, we put it in this "utility" file, instead of the "main" file.
+import 'dart:async';
+import 'dart:math';
+import 'dart:typed_data';
+
+import 'package:flutter/material.dart';
+import 'package:ten_ten_one/generated/bridge_definitions.dart';
+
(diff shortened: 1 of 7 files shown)


klochowicz (commented):
would you mind partitioning the commits into separate chunks?
an example could be:
- update Rust edition to 2021 (already merged, so it would dissapear)
- generate bip39 wallet 
- add flutter UI for the wallet
- use a stream API for updating the wallet

I understand that you want to send off something before you clock off, but then I need to spend a lot of time trying to weed through it 🙁.
 It also makes it harder to understand your thought process (what & why have you done things), especially when we work in different time zones and I can't ask for more clarification. 
Hopefully with this our async collab would be smoother 🤞

Furthermore, I see that you've squashed my commits there too - it would been easier me if could see what you've actually changed after I handed things over (even if we squash things in the end)

edit: after re-reading this comment, I have to admit that it sounded way harsher than intended (I reworded it). I'm glad that we had a chance to talk about it online and show what I meant by this, including how to partition chunks without much work.

holzeis:
closing in favor of https://github.com/itchysats/10101/pull/81

PR #76, opened 2022-10-28
Revert bogus commits

klochowicz:
I've noticed a couple of commits that raised my eyebrows. I decided to revert them, as I could not see a clear reason for their existence.
- dprint addition - we already had it, it's pointless to copy an additional one from ItchySats.
- moving ffi bindings - there's no need to depart from convention, and it raises clippy warning ("module inception"). @holzeis if I missed something in this one, please feel free to submit this one again, with some description why is this important!


klochowicz:
This cleanup was the last commit I've borsed without explicit approval. I'm sorry if I started an anti-pattern, if I hadn't merged without approvals first, @holzeis probably wouldn't either.
I hope we can all just wait for reviews now when the project is sallet up ☺️

Issue #11, opened 2022-10-17
Create wallet upon first start


holzeis:
Upon first start of the app a wallet needs to be generated utilizing the Bip39 standard. This may be implemented using a combination of the [bitcoindevkit](https://bitcoindevkit.org/) and the [lightningdevkit](https://lightningdevkit.org/)

**Acceptance Criterias**

- lightning wallet (on & off chain) is initialized and ready to use (exposed through the flutter_rust_bridge)
- a simple view showing the current balance - demonstrating that the wallet can be accessed from the view.

holzeis:
This issue is not intended to make the wallet already 100% usable, but get to a point where we feel comfortable of making it usable. Follow up issues may be required!

klochowicz:
@holzeis I've noticed we're both assigned to it - how would you like to split the work? would you like to pair on this?

bonomat:
I thought I'd share some insights: 
rust-lightning is very modular, meaning, we will need to implement various things. Among others, rust-lightning does not take care of an on-chain wallet for you but asks you to perform certain tasks and hand over the signed transactions to the ldk. 
This means, we can implement an on-chain wallet using BDK as we are used to. 

Sample code: here they use bitcoind's wallet: 

https://github.com/lightningdevkit/ldk-sample/blob/8984205acf1d328837d0f33798e89e479aea1882/src/main.rs#L147-L165

holzeis:
thanks for the input. I will have a look..

klochowicz:
@holzeis do you think we can close it? on-chain wallet is initialised, balance is displayed, and the off-chain wallet is covered by #82

holzeis:
> @holzeis do you think we can close it? on-chain wallet is initialised, balance is displayed, and the off-chain wallet is covered by #82

yes.

PR #81, opened 2022-10-28
BIP39 derived seed for a BIP84 wallet

### LINKED ISSUE #68: Create BIP-39 compatible BDK wallet
Upon the first start of the application, a new bitcoin wallet should be created from with a seed compatible with BIP39.
This should allow us to export the seed words.

holzeis:
This PR derives a BIP84 wallet from a BIP39 seed phrase and exposes a corresponding rust api through the flutter rust bridge. Additionally this PR already improves the structuring (to the best of my limited flutter skills).

resolves #68 #11 

This PR is still draft as the seed screen needs a little bit more care.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -9,9 +9,9 @@ crate-type = ["cdylib", "staticlib"] # Android needs cdylib, whereas iOS staticl
 
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
-flutter_rust_bridge = "1.49.0"
-bip39 = "1.0.1"
-rand = "^0.6.0"
-hkdf = "0.12"
-sha2 = "0.10"
 bdk = "0.23.0"
+bip39 = "1.0.1"
+flutter_rust_bridge = "1.49.0"
+hkdf = "0.12"
+rand = "^0.6.0"
+sha2 = "0.10"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -12,3 +12,6 @@ anyhow = { version = "1", features = ["backtrace"] }
 flutter_rust_bridge = "1.49.0"
 bip39 = "1.0.1"
 rand = "^0.6.0"
+hkdf = "0.12"
+sha2 = "0.10"
+bdk = "0.23.0"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,3 +10,5 @@ crate-type = ["cdylib", "staticlib"] # Android needs cdylib, whereas iOS staticl
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 flutter_rust_bridge = "1.49.0"
+bip39 = "1.0.1"
+rand = "^0.6.0"
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -1,4 +1,5 @@
 use crate::seed::Bip39Seed;
+use anyhow::bail;
 use anyhow::Result;
 use bdk::bitcoin::Network;
 use bdk::blockchain::ElectrumBlockchain;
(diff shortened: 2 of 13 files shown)


holzeis:
Actually, that PR is already quite big and there are a lot of useful things already in there. I've created a follow up ticket to build the seed widget. https://github.com/itchysats/10101/issues/84.

rust/Cargo.toml
```diff
@@ -10,3 +10,8 @@ crate-type = ["cdylib", "staticlib"] # Android needs cdylib, whereas iOS staticl
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 flutter_rust_bridge = "1.49.0"
+bip39 = "1.0.1"
```
klochowicz, line 13: I suspect you can just the version integrated with `bdk` instead by enabling the relevant feature flag? as a bonus, it enables to nicely create descriptor wallets from it (so your code would get simpler too!)

rust/Cargo.toml
```diff
@@ -10,3 +10,8 @@ crate-type = ["cdylib", "staticlib"] # Android needs cdylib, whereas iOS staticl
...
+rand = "^0.6.0"
+hkdf = "0.12"
+sha2 = "0.10"
+bdk = "0.23.0"
```
klochowicz, line 17: note to self: I forgot to enable dprint for toml files 🤦‍♂️ should be fixed in a sec :)

klochowicz (commented):
nice start! I haven't look much into flutter here, I was focusing on Rust feedback

da-kami:
Given that @klochowicz is working on figuring out what is best to achieve libaray API calls into ruest, I would propose that we don't implement complete features where we touch the UI *and* the rust backend, but we implement either UI or backend.
I think the initial focus should be on the UI, because the UI as consumer should drive the API design for what is needed in rust. I propose that we focus on the when working on issues like #85 and #84, and plug in the rust code later. We *can* discuss the API while working on such tickets and then ramp up the rust code in parallel.

da-kami:
> [snip]
> This PR is still draft as the seed screen needs a little bit more care.

It's not marked as Draft PR on GiHub, is that on purpose?

rust/src/wallet.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub const MAINNET_ELECTRUM: &str = "ssl://blockstream.info:700";
+pub const TESTNET_ELECTRUM: &str = "ssl://blockstream.info:993";
+
+#[derive(Debug)]
+pub struct WalletInfo {
+    pub phrase: Vec<String>,
+}
+
+pub fn init_wallet(network: &str) -> Result<WalletInfo> {
```
da-kami, line 21: For me it is unclear if this API is what we actually want. It feels awkward to instantiate an Electrum client on every call, I think we need different lifecycle management here. 

@klochowicz is working on the lifecycle and state management of the rust components within Dart. I would focus on the UI and not try to achieve the *actual* backend call yet but plug in dummy values for now. 

You could even go as far as implementing a value provider (e.g. factory) that can be switched from dummy value to an actual API call easily.

da-kami (approved):
Happy to merge this PR, but I am a bit confused that this is the first thing we worked on. I would have expected a PR that adds the Dashboard with dummy values similar to what is mocked in the wireframe and described in #85 and only then wire in API calls.

rust/Cargo.toml
klochowicz: this is fixed, you can rebase and run `dprint fmt` on it :)

rust/src/wallet.rs
```diff
@@ -0,0 +1,65 @@
...
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::wallet;
+
+    #[test]
+    fn init_wallet_success() {
+        wallet::init_wallet("bitcoin").expect("wallet to be initialized");
```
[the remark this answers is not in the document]
holzeis: yes. it's defined by `bdk::bitcoin::Network`

da-kami (approved):

rust/src/wallet.rs
```diff
@@ -0,0 +1,65 @@
...
+use bdk::Wallet;
+use std::str::FromStr;
+
+pub const MAINNET_ELECTRUM: &str = "ssl://blockstream.info:700";
+pub const TESTNET_ELECTRUM: &str = "ssl://blockstream.info:993";
+
+#[derive(Debug)]
+pub struct WalletInfo {
+    pub phrase: Vec<String>,
+}
```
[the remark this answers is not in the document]
holzeis: removed that code.

holzeis:
Thanks @klochowicz @da-kami  for the comments. The rest of them will be addressed in follow up PRs.

PR #91, opened 2022-10-31
Initial wallet state management

klochowicz:
See commit descriptions for details


holzeis (approved):
Thank you! LGTM 👍 

The wallet info state (app state) or similar will be managed through providers. See https://github.com/itchysats/10101/pull/94  as a reference.

https://github.com/itchysats/10101/pull/94/files#diff-e61eb31d013d12616f5532636a88cfa63631dda8f7829e5424e68542214d1608R8

rust/src/wallet.rs
```diff
@@ -1,45 +1,81 @@
...
 }
 
+pub fn get_balance() -> Result<bdk::Balance> {
+    println!("Wallet sync called");
+    WALLET
+        .try_get()
+        .context("Wallet uninitialised")?
+        .lock()
+        .map_err(|_| anyhow!("cannot acquire wallet lock"))?
+        .sync()
```
da-kami, line 76: Just out of interest, we don't have to change anything for now:

Would we be able to let rust handle the sync interval given that the wallet lives in rust? 
Then we could just read the latest balance instead of awaiting it. 

This would have the advantage that dart does not have to await the balance - potential disadvantage could be that we have to add a check for the balance to be outdated - but we need logic around that anyway, for the current code I would expect to have a timeout around sync at some point.

da-kami (approved):

rust/src/wallet.rs
klochowicz: that would be harder, but not impossible. It would require us to dedicate a thread to run tokio runtime or something and never return - will need to have a deeper look into how to even ensure this thread would not be assigned to other calls from Flutter later etc.
It will require more `Act<Mutex>` to store the value too, as you're not guaranteed that you will get the value (unless you "push" via an async stream, which I'll be trialing later on today). 

pro of the current situation (so far):
- flutter awaits, which is non-blocking, and we're only using one thread from the Rust thread poo
- no need for 2 separate async runtimes (dart + rust) - this might need to change, we'll see.

tl;dr: stay tuned, there might be changes :)

PR #95, opened 2022-11-01
Fix a few things inside wallet

klochowicz:
See commit description for details.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -3,14 +3,13 @@ use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
-use bdk::bitcoin::Network;
+use bdk::bitcoin;
 use bdk::blockchain::ElectrumBlockchain;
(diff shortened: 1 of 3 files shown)


rust/src/wallet.rs
```diff
@@ -19,23 +18,37 @@ pub const TESTNET_ELECTRUM: &str = "ssl://blockstream.info:993";
 /// Wallet has to be managed by Rust as generics are not support by frb
 static WALLET: Storage<Mutex<Wallet>> = Storage::new();
 
+pub enum Network {
+    Mainnet,
+    Testnet,
+    Regtest,
```
holzeis, line 24: ❓ Why are we adding regtest as enum option, if we do not support it?

holzeis (approved):
Nice 👍

rust/src/wallet.rs
klochowicz: I was just converting the tests initially when I added it. I anticipate it being useful when testing lightning

PR #102, opened 2022-11-01
Wire wallet seed with seed screen

### LINKED ISSUE #101: Wire wallet seed with seed screen
Currently the seed is mocked, it should be wired with the seed phrase used to generate the BIP84 wallet.

Acceptance Criteria

- Seed Phrase is fetched from the rust lib

At some point we should think about storing the seed file to the mobile device, but this is not focus of this issue.

holzeis:
resolves #101


integration_test/dashboard_test.dart
```diff
@@ -84,6 +88,8 @@ void main() {
     });
 
     testWidgets('test if seed backup warning disappears after checkbox is checked', (tester) async {
+      await api.initWallet(network: Network.Testnet);
```
da-kami, line 91: Does this mean that this test has a dependency on `testnet`, i.e. it actually requires an internet connection and fetches from a public node? 

I think it would be great if these tests would not depend on external things. Not super important for now.
On a more general note: For this test the actual seed really should not matter, why do we even need the api call here?

lib/seed.dart
```diff
@@ -18,6 +19,25 @@ class _SeedState extends State<Seed> {
...
 
+  // initialise the phrase with empty words - in order for the widget to not throw
+  // an error while waiting for the rust api. Seems to be the easiest way of handling
+  // uninitialised state, as the words will be ready immediately after the widget is
+  // initialised and the words are not visible at first.
+  List<String> phrase = ["", "", "", "", "", "", "", "", "", "", "", ""];
+
+  @override
+  void initState() {
+    _callGetSeedPhrase();
```
da-kami, line 30: ❓ Can it happen that this call never finishes or can we be sure that the seed is available through the API?

da-kami (approved):

integration_test/dashboard_test.dart
klochowicz: maybe we could run 'signet' on this? (and not connect with any external infrastructure?). Alternatively, we could allow some custom `None` network where our wallet would be hardcoded to some value and not using any public infrastructure.

lib/main.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:async';
```
klochowicz, line 1: thank you for the great commit message! I've learned something new today 👍

lib/seed.dart
```diff
@@ -18,6 +19,25 @@ class _SeedState extends State<Seed> {
   bool checked = false;
   bool visibility = false;
 
+  // initialise the phrase with empty words - in order for the widget to not throw
+  // an error while waiting for the rust api. Seems to be the easiest way of handling
+  // uninitialised state, as the words will be ready immediately after the widget is
+  // initialised and the words are not visible at first.
+  List<String> phrase = ["", "", "", "", "", "", "", "", "", "", "", ""];
```
klochowicz, line 26: I guess it's good for now, we can always refactor it later

rust/src/api.rs
```diff
@@ -16,6 +16,14 @@ pub fn init_wallet(network: Network) -> Result<()> {
...
-pub fn get_balance() -> anyhow::Result<Balance> {
+pub fn get_balance() -> Result<Balance> {
     Ok(Balance::new(wallet::get_balance()?.confirmed))
 }
+
+pub fn get_seed_phrase() -> Result<Vec<String>> {
+    let seed_phrase = wallet::get_seed_phrase()?;
+
+    // The flutter rust bridge generator unfortunately complains when wrapping a ZeroCopyBuffer with
+    // a Result. Hence we need to copy here (data isn't too big though, so that should be ok).
```
klochowicz, line 27: yeah, it's sent over once, so it's not a problem at all.

sometimes boxing (ie. wrapping in a box, like : `Result<Box<Type>>`) can help, as it just passes a pointer to  data already on the heap, but I don't think it's worth it here

rust/src/wallet.rs
```diff
@@ -27,6 +27,7 @@ pub enum Network {
 pub struct Wallet {
     blockchain: ElectrumBlockchain,
     wallet: bdk::Wallet<MemoryDatabase>,
+    seed_phrase: Vec<String>,
```
klochowicz, line 30: are you planning to address my earlier comments from the initial seed review? I would prefer if we didn't need to keep strings around, so we don't accidentally print them out in tracing logs etc.... (can be done as a follow-up, no need to block this PR)

rust/src/wallet.rs
```diff
@@ -89,6 +94,16 @@ pub fn get_balance() -> Result<bdk::Balance> {
         .sync()
 }
 
+pub fn get_seed_phrase() -> Result<Vec<String>> {
+    let wallet = WALLET
+        .try_get()
+        .context("Wallet uninitialised")?
+        .lock()
+        .map_err(|_| anyhow!("cannot acquire wallet lock"))?;
```
klochowicz, line 102: 🛠 might be a good time to extract this boilerplate into a free function

integration_test/dashboard_test.dart
holzeis: > Does this mean that this test has a dependency on `testnet`, i.e. it actually requires an internet connection and fetches from a public node?

Yes it does, but its not different for any other test that is running the TenTenOneApp, as the wallet will be initialised there.  It's only done here explicitly, because the dashboard_test are running a different app config to be able to access the providers. 

> On a more general note: For this test the actual seed really should not matter, why do we even need the api call here?

Well, because the tests are failing otherwise due to a rust error. The seed is trying to access an uninitialised wallet. Once the logging PR is merged I will adapt and add an error handling.

lib/seed.dart
holzeis: > Can it happen that this call never finishes
It could, but it wouldn't block the screen.

> can we be sure that the seed is available through the API?
No we can't, e.g. if the wallet is not initialised this call will fail. (that's the reason for the init wallet in the integration tests).

Error handling should be there, but flutter complains about simply printing the error so I opened the logging PR to be able to log errors properly. I will adapt here once https://github.com/itchysats/10101/pull/104 is merged.

rust/src/wallet.rs
holzeis: Yes 🙂 https://github.com/itchysats/10101/pull/102/commits/b21679b29dcc501066a810b0267aded26b755ab1

integration_test/dashboard_test.dart
holzeis: @da-kami https://github.com/itchysats/10101/pull/102/commits/451f23039bf77b019f914b33f5a358c2904dfda0

lib/seed.dart
holzeis: Added error handling https://github.com/itchysats/10101/pull/102/commits/451f23039bf77b019f914b33f5a358c2904dfda0

rust/src/wallet.rs
holzeis: https://github.com/itchysats/10101/pull/102/commits/510d5d4eed1bdcfd082f8217f3b0e0c1e918c0c5

rust/src/seed.rs
```diff
@@ -42,6 +36,16 @@ impl Bip39Seed {
 
         Ok(ext_priv_key)
     }
+
+    pub fn get_seed_phrase(&self) -> Vec<String> {
+        let phrase = self
+            .mnemonic
+            .to_string()
+            .split(' ')
```
klochowicz, line 44: nit: you can probably use `word_iter()` function

klochowicz (approved):
nice one! thanks for addressing the comments :-)

klochowicz:
@holzeis I hope you don't mind that I autosquashed the fixup commits to allow borsing :)

PR #104, opened 2022-11-01
Log dart messages to file

### LINKED ISSUE #103: Log dart messages to file
https://medium.com/flutter-community/logging-in-flutter-a583b3668228

holzeis:
This will allow to log messages using the FLog, which allows to export any logs to file.

resolves #103

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -18,10 +18,16 @@ import 'ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 BalanceModel balanceModel = BalanceModel();
 SeedBackupModel seedBackupModel = SeedBackupModel();
 
-void main() => runApp(MultiProvider(providers: [
-      ChangeNotifierProvider(create: (context) => balanceModel),
-      ChangeNotifierProvider(create: (context) => seedBackupModel),
-    ], child: const TenTenOneApp()));
+void main() {
+  FlutterError.onError = (details) {
+    FlutterError.presentError(details);
+  };
+
+  runApp(MultiProvider(providers: [
+    ChangeNotifierProvider(create: (context) => balanceModel),
+    ChangeNotifierProvider(create: (context) => seedBackupModel),
+  ], child: const TenTenOneApp()));
+}
 
 class TenTenOneApp extends StatefulWidget {
   const TenTenOneApp({Key? key}) : super(key: key);
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,4 +1,6 @@
+import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart' hide Size;
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:provider/provider.dart';
 import 'package:ten_ten_one/cfd_trading.dart';
 import 'package:ten_ten_one/dashboard.dart';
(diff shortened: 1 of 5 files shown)


ios/Runner/Info.plist
```diff
@@ -43,5 +43,9 @@
 	<false/>
 	<key>CADisableMinimumFrameDurationOnPhone</key>
 	<true/>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+    <string>FLogs would like to save photos from the app to your gallery</string>
+    <key>NSPhotoLibraryUsageDescription</key>
+    <string>FLogs would like to access your photo gallery for uploading images to the app</string>
```
klochowicz, line 49: 🤔 this might sound a bit fishy to the users, but indeed this is the way they recommend to use this package. I guess it just as good as anything.

lib/main.dart
```diff
@@ -32,7 +34,15 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
     super.initState();
-    _callInitWallet();
+
+    _callInitWallet()
+        .then((value) => FLog.info(text: "Successfully initialised wallet"))
+        .catchError((error) => {
+              if (error is FfiException)
+                {FLog.error(text: "Wallet failed to initialise: Error: " + error.message)}
+              else
+                {FLog.error(text: "Wallet failed to initialise: Unknown error")}
```
klochowicz, line 44: is there a way of having a generic handler that handles uncaught exceptions/errors?

klochowicz (approved):
nice one! :)

have you had any chance of testing it it whether it actually works on Android and iOS? 
It would be good to check before we go to far, because the cost of changing the logging package will just increase as we go :)

ios/Runner/Info.plist
da-kami: ❓If the user does not give the permission, the app still works right? (I hate it when apps require certain permissions and otherwise don't work even though the permissions are not actually essential for using the app...)

We could try to add a bit more context so the user understands better why this is needed and what `FLogs` is. 
Do we even have to mention `FLogs` or can we just state something along the lines of:

```
In order to save log information and debug data the app would like to save photos from the app to your gallery.
```

Why is the other direction needed - I get that we need rights to save images, but why also uploading rights?

da-kami (approved):

ios/Runner/Info.plist
klochowicz: We could also try a different avenue - AFAIK there's no need for any of these permissions on iOS if you only use the 'send logs to the devs' feature in an event of a crash. did that in testflight and app I was beta testing didn't need any permissions.

holzeis: > 🤔 this might sound a bit fishy to the users, but indeed this is the way they recommend to use this package. I guess it just as good as anything.

I just tested it on the iphone simulator and FLog.exportLogs() worked with and without the permissions set. I guess it needs to be tested on a real device?

However, I looked over the [docs](https://pub.dev/packages/f_logs) and there are also other ways of exporting the logs - no need to decide now.

I am going to remove the permissions for now and leave that decision up for https://github.com/itchysats/10101/issues/105

lib/main.dart
holzeis: Yes, https://github.com/itchysats/10101/pull/104/commits/192a5d532f6a2bde4f6c71795b25cdde7f8b543f

I've added a rudimentary implementation, but we should spent a little bit more time on how to handle errors e.g.

- automatically send the logs to us https://github.com/itchysats/10101/issues/105
- design an error screen
- etc.

PR #110, opened 2022-11-02
feat: Send Rust logs to flutter

klochowicz:
Provide some basic API to stream Rust logs to Flutter and stream existing logs
    instead of using `println!` macro.

TODO: Use `tracing` infrastructure to stream the logs, instead of handcrafted function
TODO: after adding support for `tracing` and adding a Flutter log package, send
    over timestamps, log level etc.

### DIFF
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -1,0 +1,24 @@
+use flutter_rust_bridge::StreamSink;
+use state::Storage;
+
+/// Wallet has to be managed by Rust as generics are not support by frb
+static LOG_STREAM_SINK: Storage<StreamSink<LogEntry>> = Storage::new();
+
+/// Struct to expose logs from Rust to Flutter
+pub struct LogEntry {
+    // TODO: Add more fields, including time and level
+    pub msg: String,
+}
+
+pub fn create_log_stream(sink: StreamSink<LogEntry>) {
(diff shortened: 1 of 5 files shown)


rust/src/logger.rs
```diff
@@ -0,0 +1,24 @@
+use flutter_rust_bridge::StreamSink;
+use state::Storage;
+
+/// Wallet has to be managed by Rust as generics are not support by frb
+static LOG_STREAM_SINK: Storage<StreamSink<LogEntry>> = Storage::new();
+
+/// Struct to expose logs from Rust to Flutter
+pub struct LogEntry {
+    // TODO: Add more fields, including time and level
```
holzeis, line 9: nit: The time is not needed, as the actual print will be done by f_log and this will add the time automatically. (note, there is no way to overwrite the time on the f_log messages, so the time captured here would be lost, unless we add it to the message, but I don't think that's needed)

holzeis (approved):
Thanks for picking that up 👍

rust/src/logger.rs
klochowicz: thanks for the heads up! that's the reason why I tried not to include more than necessary as I wasn't sure what's needed 😁(good ol' YAGNI)
I will clean this up when wiring it all together.

da-kami (approved):

PR #112, opened 2022-11-02
Create cfd trading offer screen

### LINKED ISSUE #87: Create cfd trading offer screens
Image

holzeis:
Simulator Screen Recording - iPhone 14 Pro Max - 2022-11-03 at 20 16 30

resolves #87


holzeis:
This PR does not include all cfd_trading screens, but I figured I already create a PR for that as there are already quite a lot of changes. Better to get it merged and create another PR for the next screens.

da-kami (approved):

da-kami:
One thing that I noticed is, that we use different names for naming the model files / classes (some of them are post-fixed with `model`, but the newly added order is not). Not a big deal, I might open a follow up PR for some alignment.

PR #117, opened 2022-11-03
Speed-dial FAB for send and receive

da-kami:
After wasting some time to work on creating an extendable FAB I noticed that there are quite a lot of libraries for this already available, so I just picked one and used it. 

We can easily change this later if we don't like it.

---

image

image

### DIFF
--- a/lib/dashboard.dart
+++ b/lib/dashboard.dart
@@ -6,9 +6,12 @@ import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/menu.dart';
 import 'package:ten_ten_one/models/seed_backup_model.dart';
 import 'package:ten_ten_one/models/service_model.dart';
+import 'package:ten_ten_one/receive.dart';
 import 'package:ten_ten_one/seed.dart';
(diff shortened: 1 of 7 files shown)


macos/Podfile.lock
```diff
@@ -1,15 +1,21 @@
 PODS:
   - FlutterMacOS (1.0.0)
+  - path_provider_macos (0.0.1):
```
da-kami, line 3: @klochowicz would you have any idea why this file and `macos/Runner.xcodeproj/project.pbxproj` was updated? It was not my intention to do so. Should these files actually be checked in?

klochowicz: I think this is just the new flutter & xcode. happened to me as well. I believe that  we should have these files checked in so that it's reproducible.

klochowicz: if possible, let's just extract them in a separate commit in this PR.

klochowicz (approved):
I'm unsure about FAB, but it does the job for now, and we can always change it later :) happy to try it out!

holzeis (approved):

PR #119, opened 2022-11-03
Payment list

da-kami:
Add a simple payment list filled with generated dummy values.
There is some room for finetuning for a better UX here, I tried to keep it simple.

image

image

--- 

Following is a list of improvements that can be done in follow-ups. 
Some of these little improvements are time-sinks so I stopped improving things for now :)

- Display the `+` sign as well (not just `-`)
- Keep the balance pinned to the top when scrolling (could also keep the service-tiles selector, but maybe shrink it upon scrolling)
- Improve the look and feel of the list items - It would be cool to also have a meaningful subtitle (e.g. the date/time when this happened...)

### DIFF
--- a/lib/dashboard.dart
+++ b/lib/dashboard.dart
@@ -7,6 +7,7 @@ import 'package:ten_ten_one/menu.dart';
 import 'package:ten_ten_one/models/seed_backup_model.dart';
 import 'package:ten_ten_one/models/service_model.dart';
 import 'package:ten_ten_one/receive.dart';
+import 'package:ten_ten_one/receive.dart';
 import 'package:ten_ten_one/seed.dart';
 import 'package:ten_ten_one/send.dart';
 import 'package:ten_ten_one/service_card.dart';
--- a/lib/dashboard.dart
+++ b/lib/dashboard.dart
@@ -6,13 +6,15 @@ import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/menu.dart';
 import 'package:ten_ten_one/models/seed_backup_model.dart';
 import 'package:ten_ten_one/models/service_model.dart';
-import 'package:ten_ten_one/receive.dart';
+import 'package:ten_ten_one/payment_history_list_item.dart';
 import 'package:ten_ten_one/receive.dart';
 import 'package:ten_ten_one/seed.dart';
 import 'package:ten_ten_one/send.dart';
 import 'package:ten_ten_one/service_card.dart';
 import 'package:ten_ten_one/utilities/divider.dart';
 import 'package:flutter_speed_dial/flutter_speed_dial.dart';
+
+import 'mocks/payment_history.dart';
 
 class Dashboard extends StatefulWidget {
   const Dashboard({Key? key}) : super(key: key);
@@ -40,6 +42,20 @@ class _DashboardState extends State<Dashboard> {
     if (!seedBackupModel.backup) {
       widgets.add(const BackupSeedCard());
     }
+
+    final paymentHistory = mockPaymentHistory(10);
+
+    final paymentHistoryList = ListView.builder(
+      shrinkWrap: true,
+      physics: const ClampingScrollPhysics(),
+      padding: const EdgeInsets.symmetric(vertical: 8.0),
+      itemCount: paymentHistory.length,
+      itemBuilder: (context, index) {
+        return PaymentHistoryListItem(data: paymentHistory[index]);
+      },
+    );
+
+    widgets.add(paymentHistoryList);
 
     return Scaffold(
       appBar: AppBar(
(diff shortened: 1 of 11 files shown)


lib/dashboard.dart
```diff
@@ -40,6 +43,20 @@ class _DashboardState extends State<Dashboard> {
       widgets.add(const BackupSeedCard());
     }
 
+    final fakeStuff = Container(
```
holzeis, line 46: 🔧 I wouldn't call the widget fakeStuff 🙈 Ideally you can separate the data from the presentation.. fake payments / data etc. is fine. But the widget should work the same with mock data as well as real data alike.. 😉 

```suggestion
    final paymentHistory = Container(
```

lib/dashboard.dart
```diff
@@ -40,6 +43,20 @@ class _DashboardState extends State<Dashboard> {
...
+        padding: const EdgeInsets.only(left: 4.0, right: 4.0),
```
holzeis, line 47: 🤔 I think the widgets should not be opinionated about their margins. If we need a margin I suggest we manage that on a top level, e.g. as padding on the ListView in the Scaffold. This way we don't have to maintain the margin for each widget individually, but only once.

lib/dashboard.dart
```diff
@@ -131,3 +148,68 @@ class BackupSeedCard extends StatelessWidget {
         ));
   }
 }
+
+@immutable
+class PaymentListItem extends StatelessWidget {
```
holzeis, line 153: 🤔 I suggest extracting that into a separate class.

lib/dashboard.dart
```diff
@@ -131,3 +148,68 @@ class BackupSeedCard extends StatelessWidget {
...
+  const PaymentListItem({
+    super.key,
+    required this.data,
+  });
+
+  final PaymentListItemData data;
```
holzeis, line 159: 🤔 I typically keep class variables above the constructor.

lib/models/amount.model.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:decimal/decimal.dart';
+
+class Amount {
+  Decimal sats = Decimal.zero;
+
+  Amount(int sats) {
+    this.sats = Decimal.fromInt(sats);
+  }
+
+  int get asSats => sats.toBigInt().toInt();
```
holzeis, line 10: ❓ what does zoBigInt().toInt() do?

lib/models/balance.model.dart
```diff
@@ -1,10 +1,10 @@
 import 'package:flutter/material.dart';
+import 'package:ten_ten_one/models/amount.model.dart';
 
 class BalanceModel extends ChangeNotifier {
-  // TODO: Change to 0, just to see that it updates :)
-  int amount = -1;
```
holzeis, line 5: I am surprised that the tests do not fail after that change.

lib/dashboard.dart
klochowicz: great idea. 👍 when the widget amount grows, things would become unwieldy.

lib/models/balance.model.dart
da-kami: That is indeed surprising! 
I played locally with it and got it failing - but only after I changed the test to use a different unexpected value (I changed it to `1000`); after that it also failed with `-1`. Afterwards changing it to `0` made it pass as expected.

I think somehow the state is not properly re-built? Do we cache things for the test?
@klochowicz any idea what could cause this?

klochowicz: I'll try to play with it to see what's happening with it.

lib/models/amount.model.dart
da-kami: It's a decimal, which is not guaranteed to fit into `int` - that's why it only offers `toBigInt`, which then offer `toInt`. I assume that the last conversion can actually fail, i.e. result in an exception. But given that we operate with `sat` amounts I don't expect them to exceed Integer's max-value; so I did not bother implementing exception handling yet 🤷‍♂️

lib/dashboard.dart
da-kami: nit: We are talking about padding here, not margins. (padding applies to the children, margin applies to the element itself)

Note: I removed the padding on the list view. Note: The `25` padding on the overall screen that was added will have to be changed to something smaller; I use the iPhoneSE to simulate the looks and I feel `25` wastes too much space. 
There are several screens that have an overflow on iPhoneSE, I will see to add a follow up that fixes this. 
In the future we can consider making certain screen properties [responsive](https://docs.flutter.dev/development/ui/layout/adaptive-responsive) to make the most out of changing screen design :)_

lib/dashboard.dart
da-kami: I suppose you mean a separate file :)
I moved it and aligned the naming a bit.

PR #121, opened 2022-11-04
chore: Add pana lint for Flutter package

klochowicz:
See: https://github.com/dart-lang/pana


bonomat:
I'll close this for now as I don't think we will go for it any time soon.

PR #125, opened 2022-11-07
Seed page layout fixes

da-kami:
In an effort to get a better hang of how flutter works and make sure our screen layout is at least somewhat responsive (the wireframes were for iPhone SE...) I worked on the layout of the seed screen.

I will see to apply my learnings to other detail / confirmation screens next :)

This is how it looks on different devices now:

image

image

---

Note: 

I went one step further and used a FAB for confirmation as well using this code:

```dart
        floatingActionButton: FloatingActionButton.extended(
            backgroundColor: checked ? Theme.of(context).colorScheme.primary : Colors.grey[300],
            foregroundColor: checked ? Colors.black : Colors.grey[600],
            onPressed: checked
                ? () {
                    final seedBackupModel = context.read<SeedBackupModel>();
                    seedBackupModel.update();
                    context.go('/');
                  }
                : null,
            label: const Text('Done'))
```

But then I read this discussion https://github.com/flutter/flutter/issues/76120 and came to the conclusion that we should not use a FAB then. I still think a FAB that can be "disabled" looks nicer, but better not go against the recommendations.

### DIFF
--- a/lib/seed.dart
+++ b/lib/seed.dart
@@ -65,103 +65,102 @@ class _SeedState extends State<Seed> {
         appBar: AppBar(
           title: const Text('Backup Seed'),
         ),
-        body: Column(
-          children: [
-            Container(
-              margin: const EdgeInsets.fromLTRB(30, 30, 30, 30),
-              child: Row(mainAxisAlignment: MainAxisAlignment.center, children: [
-                Flexible(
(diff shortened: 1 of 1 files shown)


holzeis (approved):
LGTM 👍

PR #129, opened 2022-11-07
Set log level to debug if app is not released

holzeis:


lib/main.dart
```diff
@@ -133,8 +133,13 @@ class _TenTenOneState extends State<TenTenOneApp> {
   }
 
   Future<void> _callSync() async {
-    final balance = await api.getBalance();
-    if (mounted) setState(() => balanceModel.update(Amount(balance.confirmed)));
+    try {
+      final balance = await api.getBalance();
+      balanceModel.update(Amount(balance.confirmed));
+      FLog.trace(text: 'Successfully synced wallet');
```
klochowicz, line 139: is there a way of enabling these logs? like using some sort of config file for Flutter or CLI arguments?

klochowicz (approved):
looks good, thanks :)

da-kami (approved):

lib/main.dart
holzeis: hmm.. we could use --dart-define https://itnext.io/flutter-1-17-no-more-flavors-no-more-ios-schemas-command-argument-that-solves-everything-8b145ed4285d

PR #136, opened 2022-11-07
Fix screen overflows, refactor `Amount` display

### LINKED ISSUE #127: Fix screen overflows and stick to mobile design.
On iPhone SE (3rd gen) we have several screen overflows.

We should arrange the design to be flexible enough to expand on different screens. This usually just requires a decision on what element can expand on the screen.
We can also agree to stick to a specific screen for the tournament to reduce 

Screenshots: 

image

image

---

Furthermore, our "confirmation" buttons are placed in different locations on different screens which is not a good design practice.

I looked into using FABs everywhere, but it appears that they are not meant to be disabled, see: https://github.com/flutter/flutter/issues/76120
Instead of using FABs we can make use of the `SafeArea` widget, so that the bottom of the screen area (where we usually place our button) is correctly located according to screensize.

da-kami:
image

image

image

image

resolves #127

### DIFF
--- a/lib/balance.dart
+++ b/lib/balance.dart
@@ -1,6 +1,5 @@
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
-import 'package:intl/intl.dart';
 
 import 'models/balance_model.dart';
 
@@ -9,7 +8,6 @@ class Balance extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final formatter = NumberFormat.decimalPattern('en');
     return Consumer<BalanceModel>(
       builder: (context, balance, child) {
         var amountDisplay = balance.amount.display();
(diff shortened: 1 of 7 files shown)


da-kami:
sry, screenshots bit big, I started using the simulator's quick-screenshot functionality, but that only captures the screen :D

da-kami:
Using `green` for indicating a long positions might not be a good idea, this is quite confusing...

image

Anyway, that's something that can be fixed outside this PR :)

holzeis:
> Using `green` for indicating a long positions might not be a good idea, this is quite confusing...
> 
> image
> 
> Anyway, that's something that can be fixed outside this PR :)

Yeah, but its at least consistent with the cfd offer screen 😉 

Screenshot 2022-11-07 at 12 59 43

holzeis (approved):
Thanks for fixing 👍

holzeis:
resolves #127

bonomat:
> Using `green` for indicating a long positions might not be a good idea, this is quite confusing...
> 
> image
> 
> Anyway, that's something that can be fixed outside this PR :)

The up/down arrows makes me think my position is either up or down :) Also, `Open` might be redundant given we have an `open`/`closed` separation. :)

Issue #139, opened 2022-11-08
Split wallet into `Dashboard`, `Lightning` and `Bitcoin`


da-kami:
This is necessary to depict our Bitcoin hot-wallet. In the long run we might be able to get rid of this and focus on a pure Lightning wallet, but for the current onboarding / offboarding process it is very hard to abstract this way.

Note: It might be good to have either way, this makes the distinction between Lightning and BTC clearer, one could even allow trading using the BTC wallet, but that is not modelled / planned for now.

Issue #113, opened 2022-11-02
Fund on-chain wallet


holzeis:
We need to be able to fund the on-chain wallet. A screen something like that could be implemented.

<img width="576" alt="Screenshot 2022-11-02 at 17 59 58" src="https://user-images.githubusercontent.com/382048/199557962-7f7cd6c8-12fe-4606-9937-18aff15111c4.png">

holzeis:
@da-kami any suggestions on the design?

da-kami:
For more clarity and understanding how this will work, here is what I have in my mind from the top of my head:

1. Taker -> Maker: Give me BTC address
2. Maker -> Taker: BTC address to fund
3. Taker funds BTC address with `X`
4. Maker notices funding of address with `X`
5. Maker creates initial channel with taker locking up BTC, channel capacity to be `X * 2`
6. Maker sends over `X` (minus opening fee) to Taker
7. Taker balance in 10101 lightning wallet changes.

bonomat:
What Daniel proposes sounds good. Just for clarity, he proposes an optimized approach. We could also do this:

1. Taker: internal BDK wallet: give me BTC address
	1. now the taker has money in the internal address and can also use them on-chain

If taker wants to open a lightning channel ( I assume this is the default setup)
1. Taker -> Maker: give me BTC Address but also provides details to where to open channel to (i.e. node id,)
2. Taker sends from internal wallet to maker's btc address
3. Maker creates initial channel with taker
4. Maker sends over X (minus opening fee) to Taker
5. Taker balance in 10101 lightning wallet changes.

One challenge: when the maker wants to open a channel with the taker, he needs to communicate with the taker's node, i.e. connect to an address:port. I'm not sure how we can solve this as the user's ports might not be open :/

da-kami:
This requires a view on an on-chain wallet next to the Lightning wallet in the app. I will see to extend the wireframes and then implement a simple screen for this.

da-kami:
![FundWalletJourney](https://user-images.githubusercontent.com/5557790/200468370-4b649bec-8899-4b26-ab34-94953456dac2.png)

Note: We have a Bitcoin wallet and Lightning wallet view now. I captured this in https://github.com/itchysats/10101/issues/139 - I will see to tackle https://github.com/itchysats/10101/issues/139 before this ticket (might be somewhat together).

da-kami:
[...]

PR #140, opened 2022-11-08
Rename some classes to be more expressive

da-kami:
Rename a few things in preparation for #139 - better get this in early so we don't run into friction working in parallel :)

### DIFF
--- a/integration_test/dashboard_test.dart
+++ b/integration_test/dashboard_test.dart
@@ -5,7 +5,7 @@ import 'package:integration_test/integration_test.dart';
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 import 'package:ten_ten_one/balance.dart';
-import 'package:ten_ten_one/dashboard.dart';
+import 'package:ten_ten_one/wallet_lightning.dart';
 import 'package:ten_ten_one/models/amount.model.dart';
 import 'package:ten_ten_one/models/balance_model.dart';
 import 'package:ten_ten_one/models/seed_backup_model.dart';
@@ -16,7 +16,7 @@ final GoRouter _router = GoRouter(
     GoRoute(
         path: '/',
         builder: (BuildContext context, GoRouterState state) {
-          return const Dashboard();
+          return const WalletLightning();
         },
         routes: [
           GoRoute(
@@ -29,7 +29,7 @@ final GoRouter _router = GoRouter(
   ],
 );
 
-Widget createDashboard(balanceModel, seedBackupModel) => MultiProvider(
+Widget createWallet(balanceModel, seedBackupModel) => MultiProvider(
         providers: [
           ChangeNotifierProvider<BalanceModel>(create: (context) => balanceModel),
           ChangeNotifierProvider<SeedBackupModel>(create: (context) => seedBackupModel)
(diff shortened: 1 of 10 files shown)


klochowicz (commented):
I'm not sure whether Dashboard should be the same as the whole LightningWallet screen.
I think having the ability to *not* see trading prompts for when I'm looking through my transaction history would be cool.
In other words, I'm not too keen on `Dashboard` rename.

da-kami:
> I'm not sure whether Dashboard should be the same as the whole LightningWallet screen. I think having the ability to _not_ see trading prompts for when I'm looking through my transaction history would be cool. In other words, I'm not too keen on `Dashboard` rename.

I think if we want a `Dashboard` it should be on a different level. For me `Dashboard` is an abstract word that does not add any benefit in the current context. We have a wallet view, and we are in a wallet context.

I can imagine that in the future we might have another hierarchy level on top where one can see all kind of activities of what happened - that could be a `Dashboard` then, but I don't think that's what we currently have. I think `Wallet` makes more sense in what was implemented so far - and even more with the changes outlined by #139 and #113

da-kami:
Note: For implementing #139 I will have to rename `Dashboard` to be in a lightning wallet context because there will be another view that is the Bitcoin wallet. I might be able to keep the `Dashboard` terminology as widget on top, but I don't think this makes sense in the current modularization. 

In the future we can hopefully get rid of needing a Bitcoin hot-wallet in the application, but for now we need it. cc @bonomat

klochowicz:
maybe I'll rephrase: the current widget, with big buttons about trading, is not perfect for the "wallet view" IMHO. this might be subjective, so if others I happy with the rename, I will not stop you :-)

holzeis (changes requested):
I suggest we finish this https://github.com/itchysats/10101/discussions/122 discussion, before we renaming / restructuring things.

da-kami:
> maybe I'll rephrase: the current widget, with big buttons about trading, is not perfect for the "wallet view" IMHO. this might be subjective, so if others I happy with the rename, I will not stop you :-)

I agree and recorded this observation in https://github.com/itchysats/10101/discussions/138#discussioncomment-4081307

da-kami:
> I suggest we finish this #122 discussion, before we renaming / restructuring things.

Note: @holzeis and me had a discussion including wireframe walkthrough and he will have anoher look at the changes. 
I think #122 makes sense, but I would probably see most of that discussion as post-tournament task. I don't expect to extend the POC we build, see #138 :)

da-kami:
@holzeis should this still be blocked? You re-requested a review from yourself, but it seems no review was done? 😅

integration_test/dashboard_test.dart
```diff
@@ -16,7 +16,7 @@ final GoRouter _router = GoRouter(
     GoRoute(
         path: '/',
         builder: (BuildContext context, GoRouterState state) {
-          return const Dashboard();
```
holzeis, line 19: 🤔  I guess we should rename the test file as well if its now testing the wallet lightning component instead of the dashboard.

lib/cfd_trading/cfd_offer.dart
```diff
@@ -7,7 +7,7 @@ import 'package:provider/provider.dart';
 import 'package:ten_ten_one/balance.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_order_confirmation.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
-import 'package:ten_ten_one/cfd_trading/cfd_trading_service.dart';
+import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
```
holzeis, line 10: fine with me, but for this kind of change I would have waited for the conclusion of #122.

holzeis (approved):
@da-kami feel free to merge to unblock your follow up PR. However, I would like to come back to the project structuring and naming conventions once we have concluded the discussion in #122

da-kami:
Closing in favor of https://github.com/itchysats/10101/pull/146
Sorry for the back and forth, it was not a good idea to try to do these changes separate.

PR #142, opened 2022-11-08
Use kebab-case when naming Markdown files

luckysori:
Huge nit, but we have always used this convention in the past.


bonomat (approved):
image

klochowicz (approved):

holzeis (approved):

PR #143, opened 2022-11-08
Add tracing to see Rust logs

klochowicz:
This allows us to see bdk etc logs in our app.
The logs are sent to Dart as well.

note: temporarily disable displaying debug logs from Rust in flutter (to not
      double up, esp. when tracing provides much nicer views)

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,7 +1,7 @@
 import 'dart:async';
 
 import 'package:f_logs/f_logs.dart';
-import 'package:flutter/foundation.dart' as foundation;
+// import 'package:flutter/foundation.dart' as foundation;
 import 'package:flutter/material.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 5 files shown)


lib/main.dart
```diff
@@ -32,7 +32,9 @@ void main() {
   };
 
   final config = FLog.getDefaultConfigurations();
-  config.activeLogLevel = foundation.kReleaseMode ? LogLevel.INFO : LogLevel.DEBUG;
+  // TODO: Rethink whether we need to show Rust logs from Flutter natively or
```
holzeis, line 35: 🤔 I think the switch between info and debug still makes sense also only for dart logs. If we don't want the rust logs to be shown on the flutter stdout in debug, we could reduce the level in the setup rust logging, but that may limit us in investigating issues, when a user sends us her logs (since all logs will be on the same level).

Ugly, but we could change the log level to `OFF` every time we receive a rust log in order to prevent stdout prints, save it to the database with the correct level and then changing the log level back to `DEBUG` or `INFO`

holzeis (approved):
Nice 👍

lib/main.dart
klochowicz: thanks for pointing this out, I will seek to include this comment a new cleanup ticket for flutter logs. 
We should be able to set them up from the outside ideally. Note that this is a no-op right now as there's nothing being logged on debug level (there's only trace, which was not seen anyway).

PR #146, opened 2022-11-08
Add `Dashboard`, `Lightning` and `Bitcoin` wallet view

da-kami:
resolves https://github.com/itchysats/10101/issues/139

- `Dashboard` shows the tiles that guide the user, a combined view of balance and payment history and the quick-select service tiles. There are no action buttons in this view.
- `Lightning` is the view on the lightning wallet, it shows lightning specific payment history and an expandable FAB with `receive`/`send`
- `Bitcoin` is the view on the bitcoin wallet, it shows bitcoin specific payment (i.e. transaction) history and an expandable FAB with `deposit`/`withdraw`

The `Balance` widget was extended so it is aware of Bitcoin and Lightning balance and one can decide which to display based on a selector.
The payment model was extended so it covers Bitcoin specific history as well.

Note: Channel opening is not covered in FABs within the wallets at the moment, the way to open a channel will be the `Open Channel` tile in the `Dashboard` (for now, this can be changed/extended at a later point).

---

screenshots:


da-kami:
I will re-work this and incorporate necessary renaming changes here directly. Originally I wanted to do the rename separately to agree on them faster but given the recent changes in the wireframe it will be easier to do this together with the changes around wallet / dashboard.

lib/wallet/payment_history_list_item.dart
```diff
@@ -27,6 +27,18 @@ class PaymentHistoryListItem extends StatelessWidget {
 
     String label;
     switch (data.type) {
+      case PaymentType.deposit:
```
holzeis, line 30: 🔧 On the service model we used an extensions for that. I kinda like the extension approach better, as it better hides this ugly switch 😅

lib/wallet/seed.dart
```diff
@@ -5,7 +5,7 @@ import 'package:provider/provider.dart';
 import 'package:ten_ten_one/models/seed_backup_model.dart';
 import 'package:go_router/go_router.dart';
 
-import 'ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
+import '../ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
```
holzeis, line 8: 🔧 I guess these kind of changes are the reasons for the dart style guideline not to use relative imports. 🙂

Change to `import 'package:ten_ten_one/ffi.io.dart' if (dart.libray.html) 'ffi.web.dart';`

lib/wallet/wallet.dart
```diff
@@ -0,0 +1,73 @@
...
+    WalletLightning.icon,
+    WalletBitcoin.icon,
+  ];
+
+  @override
+  Widget build(BuildContext context) {
+    WalletChangeNotifier walletChangeNotifier = context.watch<WalletChangeNotifier>();
+
+    return Scaffold(
+        appBar: AppBar(
```
holzeis, line 48: 🔧 I guess it would be nicer if the appbar is not defined by wallet, but by the wallet lightning, bitcoin or dashboard widget. That way you don't have to maintain this list of page names.

lib/wallet/wallet.dart
```diff
@@ -0,0 +1,73 @@
...
+            // title: walletChangeNotifier.selectedIndex == 0 ? const Text('Lightning Wallet') : const Text('Bitcoin Wallet'),
+            title: Text(_pageNames.elementAt(walletChangeNotifier.selectedIndex))),
+        drawer: const Menu(),
+        body: _pages.elementAt(walletChangeNotifier.selectedIndex),
+        bottomNavigationBar: BottomNavigationBar(
+          items: List.generate(
```
holzeis, line 54: ❓ why extracting the list to the top. It seems unnecessary complicated without much benefits. I suggest writing the bottom navigation bar items directly here.

Alternatively if you insist on that structure, I would at least create a new class wrapping label and icon and create a list from them. Then you can write something like this.

```
final List<Item> navigationBarItems = [...]
...
navigationBarItems.map((item) => BottomNavigationBarItem(icon: item.icon, label: item.label)).toList()
```

lib/wallet/wallet_bitcoin.dart
```diff
@@ -0,0 +1,137 @@
...
+            child: const ServiceCard(Service.cfd),
+          ),
+          const ServiceCard(Service.sportsbet),
+        ],
+      ),
+    );
+  }
+}
+
+class BackupSeedCard extends StatelessWidget {
```
holzeis, line 115: 🔧 This class is duplicated from the wallet_dashboard, please remove.

lib/wallet/wallet_bitcoin.dart
```diff
@@ -0,0 +1,137 @@
...
+            labelStyle: const TextStyle(fontSize: 18.0),
+            onTap: () => GoRouter.of(context).go(Send.route),
+          ),
+        ],
+      ),
+    );
+  }
+}
+
+class ServiceNavigation extends StatelessWidget {
```
holzeis, line 94: 🔧 This class is duplicated from the wallet_dashboard, please remove.

lib/wallet/wallet_change_notifier.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:flutter/material.dart';
+
+/// Responsible for managing the state across the different Wallet screens.
+class WalletChangeNotifier extends ChangeNotifier {
+  // the selected tab index needs to be managed in an app state as otherwise
+  // a the order confirmation screen could not change tabs to the cfd overview
```
holzeis, line 6: 🔧 Please adapt the comment.

lib/wallet/wallet.dart
da-kami: feel free to create follow up for this :)

lib/wallet/wallet.dart
da-kami: feel free to create follow up for this :)

lib/wallet/payment_history_list_item.dart
da-kami: feel free to create follow up for this :)

PR #148, opened 2022-11-08
Add confirmation to cfd settlement

### LINKED ISSUE #126: Settle confirmation screen
The settle confirmation screen is missing. It is modelled in the wireframes, but was not implemented. 

We have a confirmation screen for opening a positions, we should also have one for closing a position.
This gives the user the chance to review the action to be performed. 
Confirmation screens are good practice in financial applications.

holzeis:
resolves #126

### DIFF
--- a/lib/cfd_trading/cfd_order_detail.dart
+++ b/lib/cfd_trading/cfd_order_detail.dart
@@ -10,7 +10,7 @@ import 'package:ten_ten_one/models/order.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
 import 'package:go_router/go_router.dart';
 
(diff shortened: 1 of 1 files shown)


da-kami (approved):
LGTM! 

Gave it a try - minor thing: For me it would look nicer if the `Cancel` button would be left of the `Confirm` button the the right. But I'm not fussed about it :)

holzeis:
> LGTM!
> 
> Gave it a try - minor thing: For me it would look nicer if the `Cancel` button would be left of the `Confirm` button the the right. But I'm not fussed about it :)

I am going to merge this, feel free to open a follow up PR to switch the orders ;)

PR #155, opened 2022-11-09
Rust log improvements

klochowicz:
- fix a problem (potential multiple inits of tracing)
- address @holzeis comment from previous review (reenable Flutter debug logs)
- only send logs to Dart in release mode (we don't need them in debug, as we see them on stdout)

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -146,7 +146,11 @@ class _TenTenOneState extends State<TenTenOneApp> {
 
   Future<void> setupRustLogging() async {
     api.initLogging().listen((event) {
-      FLog.logThis(text: 'log from rust: ${event.msg}', type: LogLevel.DEBUG);
+      // Only log to Dart file in release mode - in debug mode it's easier to
+      // use stdout
+      if (foundation.kReleaseMode) {
+        FLog.logThis(text: '${event.target}: ${event.msg}', type: LogLevel.DEBUG);
+      }
     });
   }
 }
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,7 +1,7 @@
 import 'dart:async';
 
 import 'package:f_logs/f_logs.dart';
-// import 'package:flutter/foundation.dart' as foundation;
+import 'package:flutter/foundation.dart' as foundation;
 import 'package:path_provider/path_provider.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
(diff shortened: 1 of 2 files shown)


lib/main.dart
```diff
@@ -144,7 +144,11 @@ class _TenTenOneState extends State<TenTenOneApp> {
 
   Future<void> setupRustLogging() async {
     api.initLogging().listen((event) {
-      FLog.logThis(text: 'log from rust: ${event.msg}', type: LogLevel.DEBUG);
+      // Only log to Dart file in release mode - in debug mode it's easier to
+      // use stdout
+      if (foundation.kReleaseMode) {
+      FLog.logThis(text: '${event.target}: ${event.msg}', type: LogLevel.DEBUG);
```
holzeis, line 150: ❓ have you tried setting the log level off an on again for the rust logs? that way the logs would still end up in the database and are not printed to stdout. I think the level would still be helpful when reviewing the logs.

holzeis (approved):
Thanks for addressing my comments 👍

rust/src/logger.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::Once;
+
```
luckysori, line 2: 🙃 Unwanted newline I suppose.

lib/main.dart
```diff
@@ -33,9 +33,7 @@ void main() {
   };
 
   final config = FLog.getDefaultConfigurations();
-  // TODO: Rethink whether we need to show Rust logs from Flutter natively or
-  // via Flutter
-  // config.activeLogLevel = foundation.kReleaseMode ? LogLevel.INFO : LogLevel.DEBUG;
+  config.activeLogLevel = foundation.kReleaseMode ? LogLevel.INFO : LogLevel.DEBUG;
```
luckysori, line 36: ❓ Commit message says `trace` but I see `info` and `debug` here. Is that intentional?

luckysori (approved):

lib/main.dart
klochowicz: feel free to submit a follow-up; I did only the amount I needed to get me unblocked :)

rust/src/logger.rs
klochowicz: true. after tournament I might start a discussion that it would be actually nice to separate `std` imports from external crates, but it's not a good time to do now :-)

lib/main.dart
klochowicz: reworded the commit message, thanks for spotting!

PR #157, opened 2022-11-09
feat: Persist wallet seed on disk

klochowicz:
The same seed is persisted across multiple runs of the application.

XXX: This seed is not encrypted on disk, and should as such not be used in production.


holzeis (approved):
LGTM 👍

da-kami (approved):

PR #163, opened 2022-11-10
Add LDK

klochowicz:
- Use bdk-ldk crate to create a BdkLdkWallet
- setup lightning node

So far the sync on lightning happens periodically, but there's not much to see when there are no
peers connected.
You can also see outputs in `logs.txt` as well.

The PR is still unfinished, notably we should add the remaining steps from https://github.com/klochowicz/ldk-bdk-sample
(that repo is already upgraded to the current versions of `rust-lighting` and `bdk`).

- [x] add networking from bdk-ldk-sample 
- [ ] add handle_ldk_events loop from bdk-ldk-sample
- [x] add other steps needed to run lightning network

We can easily start adding the API in the subsequent PR, by picking from `cli.rs` the parts that we need and just exposing the rust API. 

it notably fails on clippy job as I've copied across things from the sample in anticipation of filling in the gaps.


klochowicz:
the good news is that it compiles for Android :-)

klochowicz:
Closed in favour of #169 (cleaned up this PR, rebased, and make it mergeable)

PR #164, opened 2022-11-10
fix: Do not error if the seed path already exists

klochowicz:
Only try to create the missing directory if it's not there already.


klochowicz:
pretty annoying regression caused by me earlier, sorry if it caused any trouble! 😅

holzeis (approved):

Issue #166, opened 2022-11-10
Replace bitcoin address in frontend with new one once the `latest` one was used


da-kami:
Not sure what is the best trigger for this, might be easiest to just do it periodically so the address in the frontend gets replaced once it was used.

luckysori:
Is this important for the tournament. AFAIK reusing addresses is a bad practice, but it still "works".

bonomat:
I don't think so, I'll remove it.

Issue #165, opened 2022-11-10
Payment history


da-kami:
Fetch Bitcoin and Lightning payment history from the wallet backend and plug it in instead of the mocked data.

This concerns the `Dashboard` where we have a combined history of Bitcoin and Lightning, and the specific Lightning and Bitcoin wallet screens.

Furthermore this payment history could serve as the basis for the CFD trading overview (which is basically a subset of the Lightning history + some additional CFD specific data).

da-kami:
@klochowicz the `LightningSystem` fields

```
    pub inbound_payments: PaymentInfoStorage,
    pub outbound_payments: PaymentInfoStorage,
```

represent Lightning payments and don't have anything to do with the on-chain wallet, correct?
I am planning to use this as basis for the lightning payment history. 

I think we will have to add a database to store additional metadata so we can display what a payment represents in context.
There is no mechanism for this, correct?

da-kami:
After discussion with @holzeis this is what I have planned for Lightning payments as part of this ticket:

1. Add additional context data for the payments. This goes towards a the model to depict / persist the CFDs. For the beginning a simple `Hashmap<PaymentHash, PaymentContext>` should do, where `PaymentContext` defines the context of the payment.
1. Expose simple payments API based on `LightningSystem` `inbound_payments` and `outbound_payments`
    1. This will give us a general "Payments" API; I would already cater for the context to be added later, but default it to `Payment Sent` and `Payment Received` (or similar)

Based on the sessions for integrating the protocol with @bonomat and @luckysori we can then ramp up the CFD model (which will likely have a separate identifier) and relate it to the payment `PaymentContext` :)

PR #167, opened 2022-11-10
Deposit bitcoin

da-kami:
image

image

image

--- 

To be done in a follow-up PR:

- Implement actual the payment history: https://github.com/itchysats/10101/issues/165
- Implement re-loading new address once used: https://github.com/itchysats/10101/issues/166


lib/wallet/depsoit.dart
```diff
@@ -0,0 +1,107 @@
+import 'package:f_logs/f_logs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
+import 'package:qr_flutter/qr_flutter.dart';
+
+import '../ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
```
holzeis, line 7: nit: Can you change that to a package import?

lib/wallet/wallet_bitcoin.dart
```diff
@@ -5,12 +5,13 @@ import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
...
 import 'package:ten_ten_one/wallet/payment_history_list_item.dart';
-import 'package:ten_ten_one/wallet/receive.dart';
 import 'package:ten_ten_one/wallet/seed.dart';
-import 'package:ten_ten_one/wallet/send.dart';
 import 'package:ten_ten_one/wallet/service_card.dart';
 import 'package:ten_ten_one/utilities/divider.dart';
 import 'package:flutter_speed_dial/flutter_speed_dial.dart';
+import 'package:ten_ten_one/wallet/withdraw.dart';
+
+import 'depsoit.dart';
```
holzeis, line 14: typo: dep**so**it.dart --> dep**os**it.dart

Also can you please change to a package import 🙂

rust/src/api.rs
```diff
@@ -9,12 +9,22 @@ pub struct Balance {
     pub confirmed: u64,
 }
 
+pub struct Address {
```
holzeis, line 12: ❓ Why do we need to wrap a single attribute into a struct. Can't we just return a String in the api?

holzeis (approved):
LGTM 👍

rust/src/api.rs
da-kami: Usually I would argue: It's not bad to wrap it in terms of extendability of the API without breaking it. If we e.g. want to add to expose a `network` field as part of the `get_address` API we can just add it without breaking changes when having an `Address` type. 

For the moment it really does not matter and returning a String would be fine - but it did not take much effort wrapping it with a type :D

PR #176, opened 2022-11-13
fix: Add values to log messages

holzeis:
Adding a vistor implementation taken from https://burgers.io/custom-logging-in-rust-using-tracing.

@klochowicz I wasn't able to enable the rust logs, so I enabled the dart logs and noticed that the piped log messages from rust only contained the field names but not the values. So I followed the blog post mentioned above to add the values.

### DIFF
--- a/rust/src/logger.rs
+++ b/rust/src/logger.rs
@@ -2,6 +2,7 @@ use anyhow::Context;
 use anyhow::Result;
 use flutter_rust_bridge::StreamSink;
 use state::Storage;
+use std::collections::BTreeMap;
 use std::sync::Once;
 use time::macros::format_description;
 use tracing_subscriber::filter::Directive;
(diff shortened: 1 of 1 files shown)


klochowicz:
@holzeis so Rust logs are not working at all for you? or is it just the level tweaking you're mentioning?

klochowicz (changes requested):
the new changeset results in logs looking like this: 

```
flutter: {_TenTenOneState} {setupRustLogging} {log: {"fields":{"log.file":"/Users/mariusz/.cargo/registry/src/github.com-1ecc6299db9ec823/rustls-0.20.7/src/client/tls13.rs","log.line":1050,"log.module_path":"rustls::client::tls13","log.target":"rustls::client::tls13","message":"Ticket saved"},"name":"log event"}} {LogLevel.DEBUG} {14 November 2022 11:24:47 AM}
```
the json serialisation seems excessive for our use-case.
Perhaps amending `LogEntry` struct to pass the additional fields would help in your use-case? In particular, tracing event fields might be useful indeed.

Also, what would you like to send through except the already present ones? the current code in main sends:
- message
- target
- level

holzeis:
> @holzeis so Rust logs are not working at all for you? or is it just the level tweaking you're mentioning?

@klochowicz unless I don't have to run flutter somehow differently, no rust logs are shown on main for me..

> the new changeset results in logs looking like this:
> 
> ```
> flutter: {_TenTenOneState} {setupRustLogging} {log: {"fields":{"log.file":"/Users/mariusz/.cargo/registry/src/github.com-1ecc6299db9ec823/rustls-0.20.7/src/client/tls13.rs","log.line":1050,"log.module_path":"rustls::client::tls13","log.target":"rustls::client::tls13","message":"Ticket saved"},"name":"log event"}} {LogLevel.DEBUG} {14 November 2022 11:24:47 AM}
> ```

Well f_log still does not provide a nice log formatting, but at least we have the log messages including the values in the database, which we hadn't before.

The rust logs on main currently look like this:

```
flutter: {_TenTenOneState} {setupRustLogging} {log: log event: {message, log.target, log.module_path, log.file, log.line}} {LogLevel.DEBUG} {14 November 2022 06:27:25 AM}
```


> the json serialisation seems excessive for our use-case. 
Fair, I removed the dependency here 102aaf6ffb108eaa07ab6e031e4d1a3706113e86


> Perhaps amending `LogEntry` struct to pass the additional fields would help in your use-case? In particular, tracing event fields might be useful indeed.
> 
> Also, what would you like to send through except the already present ones? the current code in main sends:
> 
>     * message
>     * target
>     * level

I am fine with the current set of fields. It wasn't my intention to change the logging, but rater add the values to the current logs. Just logging the name of an attribute is not suffcient - as we would have never seen any rust log values in dart. (only message or any other variable that is added to tracing)

rust/src/logger.rs
```diff
@@ -54,21 +55,65 @@ where
...
                 msg,
-                target: metadata.target().to_string(),
-                level: metadata.level().to_string(),
+                target: event.metadata().target().to_string(),
+                level: event.metadata().level().to_string(),
             });
     }
 }
 
+struct JsonVisitor<'a>(&'a mut BTreeMap<String, String>);
```
klochowicz, line 80: nit
```suggestion
struct Visitor<'a>(&'a mut BTreeMap<String, String>);
```
(here and elsewhere - we don't use json anymore)

klochowicz (approved):
thanks for your responses that helped me understand the problem; and, most of all, for fixing the issue :) LGTM!

PR #177, opened 2022-11-14
Add maker binary

klochowicz:
First cut of the binary. So far it just initialises the tracing and wallet (and then exits).
Posting now to make sure we're on the same page with the structure.


klochowicz:
note: I have found a way to make the reuse least impactful for us (no changes from the perspective of rust lib, other than making some modules public; so no rebase conflict risk).

holzeis (approved):
Thanks 👍

klochowicz:
merging this, so we can unlock all the other tickets depending on having the binary.

PR #178, opened 2022-11-14
chore: Split initialisation and running of LDK

klochowicz:
Without this, we are susceptible to a race condition (we need to *hope* that bdk
init will happen before we start other tasks depending on it). If we explicitly
wait for init first, we don't have such a problem.

Add `Clone` trait implementation to Wallet, so that we don't need to hold the mutex to the
wallet when passing some of its data around.
Note: cloning is pretty cheap here, because we're cloning only a bunch of Arc
pointers, not the underlying wallet data.

This makes the API more discoverable, allowing for the usage
`wallet.run_ldk(port)`.

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -28,11 +28,19 @@ impl Address {
 }
 
 pub fn init_wallet(network: Network, path: String) -> Result<()> {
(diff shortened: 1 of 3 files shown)


holzeis (approved):
Nice 👍

PR #179, opened 2022-11-14
Bitcoin payment history instead of mock data

da-kami:
first part of https://github.com/itchysats/10101/issues/165

Given that this is independent of the Lightning payments (to some extent), it would be great if we can merge this already :)

--- 
Bitcoin history screenshots:

image

image

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -10,6 +10,8 @@ import 'package:ten_ten_one/bridge_generated/bridge_definitions.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_order_confirmation.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_order_detail.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
+import 'package:ten_ten_one/models/payment.model.dart';
+import 'package:ten_ten_one/payment_history_change_notifier.dart';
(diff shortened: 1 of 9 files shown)


lib/main.dart
```diff
@@ -27,9 +27,9 @@ import 'package:ten_ten_one/wallet/withdraw.dart';
 
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 
-LightningBalance lightningBalanceModel = LightningBalance();
-BitcoinBalance bitcoinBalanceModel = BitcoinBalance();
-SeedBackupModel seedBackupModel = SeedBackupModel();
+LightningBalance lightningBalance = LightningBalance();
```
holzeis, line 32: ❓ should we add the suffix ChangeNotifier then? similar to the `CfdTradingChangeNotifier` and `WalletChangeNotifier`?

holzeis (approved):
LGTM 👍

lib/main.dart
da-kami: I think we should, and we could actually move them somewhere else - those are (part of) the controller for me, not the model. 

This requires a bit more cleanup because what is currently `balance_model` is actually controllers. I will see to address this in a follow up!

PR #187, opened 2022-11-15
Feat/maker ldk

da-kami:


klochowicz (approved):
contents looks good.
nit: the commits might as well be squashed together (except using regtest and nigiri, the first one), because the split doesn't work too well (there's a few unrelated changes, and then a commit called "fix formatting").

rust/src/lightning.rs
```diff
@@ -525,6 +528,110 @@ pub async fn run_ldk(system: &LightningSystem) -> Result<BackgroundProcessor> {
     Ok(background_processor)
 }
 
+pub async fn run_ldk_server(
```
klochowicz, line 531: I suspect we could just use `run_ldk` on both sides (and just listen for incoming connections on both). Two different functions, doing almost the same, make things more confusing than actually solve.

klochowicz: in the meantime, to make things more maintainable, I just refactored the code a bit (so it calls common path afterwards) - this should make it much less confusing.

rust/src/lightning.rs
```diff
@@ -558,77 +558,7 @@ pub async fn run_ldk_server(
...
-                        &PeerInfo { pubkey, peer_addr },
-                        system.peer_manager.clone(),
-                    )
-                    .await;
-                });
-            }
-        }
-    }
-    tracing::info!("Lightning network started");
+    let background_processor = run_ldk(system).await?;
```
da-kami, line 561: We could extract the listening part into a separate function and have two calls on the server, one for starting the listening and then just `run_ldk`. The `tcp_handle is not used in `run_ldk`.
Not important enough to fix now though :)

PR #193, opened 2022-11-15
feat: Integrate LDK logs into tracing

klochowicz:
These logs are  by default stored in `logs.txt`, which makes it hard to debug
during program execution due to lack of surrounding context - enabling them
in tracing makes it easier to spot problems.

Add a custom target `ldk`, so that they can be disabled easily if they become a nuisance.

Note: Unfortunately, rust-lightning is tailored to work with `log` crate instead of
`tracing`, and it's not trivial.
Whilst it can be integrated, it can be a time-sink. Meanwhile, this commit should be
sufficient for debugging purposes.


klochowicz:
not happy with this PR, but I don't want to spend more time than absolutely necessary on this until after the tournament.

holzeis (approved):

PR #192, opened 2022-11-15
Add get offers HTTP API to maker

klochowicz:
- Start Rocket with a simple GET call to get the latest offers

PR #194, opened 2022-11-15
Allow maker to force-close first channel

luckysori:
Useful for resetting the state when testing locally.

### DIFF
--- a/rust/src/lightning.rs
+++ b/rust/src/lightning.rs
@@ -171,6 +171,26 @@ pub async fn open_channel(
     Ok(())
 }
 
+/// Force-closes the _first_ channel with a peer.
+pub async fn force_close_channel(
+    channel_manager: Arc<ChannelManager>,
+    remote_node_id: PublicKey,
+) -> Result<()> {
+    let channel_list = channel_manager.list_channels();
+    let channel = channel_list
+        .iter()
+        .find(|details| details.counterparty.node_id == remote_node_id)
+        .context("No channel with peer")?;
+
+    channel_manager
+        .force_close_broadcasting_latest_txn(&channel.channel_id, &channel.counterparty.node_id)
+        .map_err(|e| anyhow!("Could not force-close channel: {e:?}"))?;
+
+    tracing::info!("Channel has been successfully force-closed");
+
+    Ok(())
+}
+
 pub(crate) enum HTLCStatus {
     // Pending, FIXME: Never constructed in the example
     Succeeded,
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -306,6 +306,18 @@ pub async fn open_cfd(taker_amount: u64, maker_amount: u64) -> Result<()> {
     Ok(())
 }
 
+pub async fn force_close_channel(remote_node_id: PublicKey) -> Result<()> {
+    let channel_manager = {
+        let lightning = &get_wallet()?.lightning;
+
+        lightning.channel_manager.clone()
+    };
+
+    lightning::force_close_channel(channel_manager, remote_node_id).await?;
+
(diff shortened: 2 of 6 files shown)


klochowicz (commented):
it would be nice if this behaviour did not break the maker API (maybe a flag instead of a command?)

alternatively, you can just create a `PUT` request via the HTTP API on top of #192

rust/src/lightning.rs
```diff
@@ -155,6 +155,20 @@ pub async fn open_channel(
     Ok(())
 }
 
+/// Force-closes the _first_ channel with a peer.
+pub async fn force_close_channel(channel_manager: Arc<ChannelManager>) -> Result<()> {
+    let channel_list = channel_manager.list_channels();
+    let channel = channel_list.first().unwrap();
```
klochowicz, line 161:
```suggestion
    let channel = channel_list.first().context("No channels in the list")?;
```

rust/src/wallet.rs
```diff
@@ -235,6 +235,18 @@ pub async fn open_channel(peer_info: PeerInfo, channel_amount_sat: u64) -> Resul
     .await
 }
 
+pub async fn force_close_channel() -> Result<()> {
+    let channel_manager = {
+        let lightning = &get_wallet()?.lightning;
+
+        lightning.channel_manager.clone()
+    };
```
klochowicz, line 243: 🙃 I guess this can be a one-liner?
```suggestion
    let channel_manager =  get_wallet()?.lightning.channel_manager.clone();
```

luckysori: Yep! I copy-pasted it from somewhere where dropping the `MutexGuard` was important. But it isn't the case here.

klochowicz (approved):
awesome! <3

rust/src/wallet.rs
luckysori: Ah, NVM. This is the problem:

```
    Checking ten_ten_one v0.1.0 (/home/luckysori/github/itchysats/10101/rust)
warning: this `MutexGuard` is held across an `await` point
   --> rust/src/wallet.rs:310:28
    |
310 |     let channel_manager = &get_wallet()?.lightning.channel_manager;
    |                            ^^^^^^^^^^^^^
    |
    = note: `#[warn(clippy::await_holding_lock)]` on by default
    = help: consider using an async-aware `Mutex` type or ensuring the `MutexGuard` is dropped before calling await
note: these are all the `await` points this lock is held through
   --> rust/src/wallet.rs:310:5
    |
310 | /     let channel_manager = &get_wallet()?.lightning.channel_manager;
311 | |
312 | |     lightning::force_close_channel(channel_manager.clone(), remote_node_id).await?;
313 | |
314 | |     Ok(())
315 | | }
    | |_^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock
```

klochowicz: hmm... I think my suggestion had a  `.clone()`, which should not trigger the warning

luckysori: Still doesn't work with `.clone()` on a separate line.

The only alternative fix that I have found is:

```rust
let channel_manager = std::sync::Arc::clone(&get_wallet()?.lightning.channel_manager);
lightning::force_close_channel(channel_manager, remote_node_id).await?;
```

I think this is because it is equivalent to:

```rust
let channel_manager = std::sync::Arc::clone({ &get_wallet()?.lightning.channel_manager });
lightning::force_close_channel(channel_manager, remote_node_id).await?;
```

Or maybe the whole expression is moved into the function?

Either way the key is to not hold the `MutexGuard` across the await point.

PR #195, opened 2022-11-15
chore: add some more beautiful icons

bonomat:
Some more beautiful icons :) 
Icons are debatable. If the tx/action is pending, the icon will be a circular loading thingy.

image

image

and after adding amount item: 

image


bonomat:
Merged by default. It's been open since 2 days without reviews and rebased a couple of times. Feel free to still have a look and we can fix things in follow-up PRs.

bors r+

Issue #182, opened 2022-11-14
Open CFD


holzeis:
Wiring the frontend/taker with the maker, adding a custom output for opening a cfd (dlc) position.

luckysori:
Done with #196.

PR #196, opened 2022-11-15
Allow taker to open a CFD

luckysori:
As far as I can tell, the taker is able to open a CFD.

### DIFF
--- a/lib/cfd_trading/cfd_order_confirmation.dart
+++ b/lib/cfd_trading/cfd_order_confirmation.dart
@@ -1,6 +1,6 @@
-import 'dart:async';
-
+import 'package:f_logs/model/flog/flog.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:go_router/go_router.dart';
 import 'package:intl/intl.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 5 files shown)


rust/src/lightning.rs
```diff
@@ -488,28 +454,62 @@ pub fn setup(
...
+        channel_manager: system.channel_manager.clone(),
+        wallet: system.wallet.clone(),
+        network_graph: system.network_graph.clone(),
+        keys_manager: system.keys_manager.clone(),
+        inbound_payments: system.inbound_payments.clone(),
+        outbound_payments: system.outbound_payments.clone(),
+        network: system.network,
+    };
+
+    // Step 17: Initialize routing ProbabilisticScorer
```
bonomat, line 478: Does `Step17` this still make sense?

da-kami (approved):
Noice! 

@luckysori it would be sweet if you could fix the copy-pasters I introduced before merging 😬

The TODOs related to persisting CFD state on taker and maker were captured in:

https://github.com/itchysats/10101/issues/201
https://github.com/itchysats/10101/issues/203

lib/cfd_trading/cfd_order_confirmation.dart
```diff
@@ -66,24 +70,39 @@ class CfdOrderConfirmation extends StatelessWidget {
...
-                                // todo send order to maker.
+
+                                // TODO: Plug in actual state changes
 
                                 // clear draft order from cfd service state
                                 cfdTradingService.draftOrder = null;
 
+                                try {
+                                  // TODO: Don't hardcode the values
+                                  await api.openCfd(takerAmount: 20000, leverage: 2);
```
da-kami, line 84: Not sure if we should keep the hardcoded value or use the actual margin. I would vote for using the actual margin but keep the leverage locked in at 2 until we action #198 

I also captured https://github.com/itchysats/10101/issues/200 which is related.

@luckysori if we keep the hardcoded value please add a comment to https://github.com/itchysats/10101/issues/200 :)

luckysori: Let's keep it hard-coded until we action #200. I'll add the comment.

da-kami:
resolves https://github.com/itchysats/10101/issues/182

(follow up tickets captured)

PR #197, opened 2022-11-15
Fetch bid ask index price from bitmex

### LINKED ISSUE #106: Subscribe to maker to fetch btcusd bid, ask and index price

holzeis:
@bonomat @da-kami Please check the spread calculation on the long and short offers. I tried to align with the logic I found in the maker automation, but I don't think I got it completely right.

- subscribes to the bitmex api for btxusd prices using the instrument target (as this is including bid, ask and index price in one message)
- skipped ethusd for now, as it would only introduce additional complexities which we don't need now.
- reused the bitmex-stream lib from itchysats
- wires offers to the cfd offer screen through periodical rust taker api calls.
- hide material app until ready

Screenshot 2022-11-16 at 10 27 26

see commit messages for more details

resolves #106

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -65,6 +65,8 @@ class TenTenOneApp extends StatefulWidget {
 }
 
 class _TenTenOneState extends State<TenTenOneApp> {
+  bool ready = false;
+
   @override
   void initState() {
     super.initState();
@@ -73,10 +75,13 @@ class _TenTenOneState extends State<TenTenOneApp> {
 
   @override
   Widget build(BuildContext context) {
-    return MaterialApp.router(
-      title: 'TenTenOne',
-      theme: ThemeData(primarySwatch: Colors.orange),
-      routerConfig: _router,
+    return Visibility(
+      visible: ready,
+      child: MaterialApp.router(
+        title: 'TenTenOne',
+        theme: ThemeData(primarySwatch: Colors.orange),
+        routerConfig: _router,
+      ),
     );
   }
 
@@ -160,16 +165,22 @@ class _TenTenOneState extends State<TenTenOneApp> {
           .then((value) => FLog.info(text: "ldk node stopped."))
           .catchError((error) => FLog.error(text: "ldk stopped with an error", exception: error));
 
-      FLog.info(text: "Successfully initialised wallet");
-    } on FfiException catch (error) {
-      FLog.error(text: "Wallet failed to initialise: Error: " + error.message, exception: error);
-    } catch (error) {
-      FLog.error(text: "Wallet failed to initialise: Unknown error");
-    }
+      final offer = await api.getOffer();
+      cfdOffersChangeNotifier.update(offer);
 
-    // initial sync
-    _callSync();
-    _callSyncPaymentHistory();
+      // initial sync
+      _callSync();
+      _callSyncPaymentHistory();
+
+      FLog.info(text: "TenTenOne is ready!");
+      setState(() {
+        ready = true;
+      });
+    } on FfiException catch (error) {
+      FLog.error(text: "Failed to initialise: Error: " + error.message, exception: error);
+    } catch (error) {
+      FLog.error(text: "Failed to initialise: Unknown error");
+    }
 
     // consecutive syncs
     runPeriodically(_callSync);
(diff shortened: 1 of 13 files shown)


maker/src/bitmex.rs
```diff
@@ -0,0 +1,161 @@
...
+        while let Some(text) = stream.try_next().await.expect("message from bitmex") {
+            tracing::trace!(%text, "Received message from bitmex");
+            let quote = Quote::from_str(&text).expect("quote from bitmex");
+            if let Some(quote) = quote {
+                tracing::debug!(
+                    "Received quote update for {:?}, bid: {}, ask: {}, index: {}",
+                    quote.symbol,
+                    quote.bid,
+                    quote.ask,
+                    quote.index,
```
bonomat, line 29: You could also log the timestamp we get from BitMEX.

maker/src/routes.rs
```diff
@@ -1,52 +1,70 @@
...
+) -> Result<Json<Offers>, HttpApiProblem> {
+    let rx_quote_receiver = rx_quote_receiver.inner().clone();
+    let quote = *rx_quote_receiver.borrow();
 
-    let long = Offer {
-        bid: 19000.into(),
-        ask: 21000.into(),
-    };
+    // TODO: take spread from clap
+    let spread = dec!(0.015);
```
bonomat, line 35: Given we don't run this in production we can remove or reduce it to a minimum :)

maker/src/routes.rs
```diff
@@ -1,52 +1,70 @@
...
+            };
 
-    let index_price = 19750.into();
+            let short = Offer {
+                bid: quote.bid.round().to_u64().expect("to fit into u64"),
+                ask: (quote.ask * (Decimal::ONE + spread))
+                    .round()
+                    .to_u64()
+                    .expect("to fit into u64"),
+            };
```
bonomat, line 53: Why do we have 2 offers each with it's own bid/ask pair?

maker/src/routes.rs
```diff
@@ -1,52 +1,70 @@
...
+use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
+use tokio::sync::watch;
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Offer {
-    bid: Price,
-    ask: Price,
+    bid: u64,
+    ask: u64,
```
bonomat, line 17: I think this should a `Decimal` or `f64`

lib/main.dart
```diff
@@ -184,6 +188,17 @@ class _TenTenOneState extends State<TenTenOneApp> {
     }
   }
 
+  Future<void> _callGetOffers() async {
+    try {
+      // TODO: the endpoint should be configurable, probably in the settings?
```
bonomat, line 193: Let's make sure to cover this in a ticket after this gets merged. It's easy to overlook otherwise.

bonomat (commented):
Thanks. I would model the offers differently and just go for a single `Offer` which has a `bid` and `ask`. This makes things easier.

maker/src/routes.rs
da-kami: I think it has to be either one offer with bid and ask, that is then mapped to long and short, or two offers with one price where one offer representing short and one long. The later is what we had in ItchySats. In ItchySats we opted for two offers because we had other parameters that would depend on the position (long/short); but the model was always confusing.

As @bonomat has stated in https://github.com/itchysats/10101/pull/197#pullrequestreview-1181644762 it might be easier to have one offer with bid and ask price for now, and then the UI derives what to display from that offer, i.e. we have one offer with the maker's bid and ask price which represents what the maker is offering to the taker - i.e. the bid represents the price for going short (selling) and the ask for going long (buying). @bonomat please validate that this makes sense ;)

maker/src/routes.rs
da-kami: We did already introduce `Decimal` for the amount type, I think we should stick to `Decimal.

maker/src/routes.rs
da-kami: One more note for clarification:

image

Bid and ask represent short and long and should not change when we switch between long and short in the UI. In ItchySats the price changed because it was part of the switch, i.e. we never displayed both prices in ItchySats but either the one for going long or short.

lib/main.dart
da-kami: I think we need this more globally because I am also working on adding calls to the maker's API.

I opened https://github.com/itchysats/10101/pull/209 for now to align this in once place.

da-kami: Here's what we came up with, with this you won't need to pass the endpoint via the API, but can just use it inside the rust code:
https://github.com/itchysats/10101/pull/210

maker/src/routes.rs
holzeis: fixed 3a6fbc18afb3b6de993197f8a35eec8cfc698d08 (maker) and 4c34e9813bfceeb710430ccbf3fd7780978eaa41 (taker)

Decimal can not be used for the taker though as Decimal does not implement the IntoDart trait. We could look into it eventually, but not now.

maker/src/routes.rs
holzeis: fixed 4c34e9813bfceeb710430ccbf3fd7780978eaa41

lib/main.dart
holzeis: Extracted into a const in rust.

https://github.com/itchysats/10101/pull/197/commits/4c34e9813bfceeb710430ccbf3fd7780978eaa41#diff-45f0b65e906e22c7ca591e0d912bc35123341f4ea5c50e5207ae63b81d62b054R8

maker/src/bitmex.rs
holzeis: fixed f533eb99bafbbd8f6381ff23a2b644e915df0c9d

maker/src/routes.rs
```diff
@@ -1,56 +1,46 @@
...
-}
-
-#[derive(Serialize, Deserialize, Debug)]
-pub struct Offers {
-    long: Offer,
-    short: Offer,
-    index_price: Price,
+    bid: Decimal,
+    ask: Decimal,
+    index: Decimal,
```
bonomat, line 18: Does this work or do we need to annotate it with how to serialize like you did below?

bonomat (approved):
Looks good, added one question, but don't bother if it works as is.

maker/src/routes.rs
holzeis: it works as it is serializing the decimal to a string and then i am parsing it from string into a float. It's probably nicer to serialize it as float. good catch - i am going to adapt.

PR #205, opened 2022-11-16
feat: Publish maker as a Docker container

klochowicz:
Based on itchysats workflow, I only removed arm builds and taker binary.
We can refactor the code later, as we probably don't need it to be as flexible anymore.

### DIFF
--- a/.github/actions/build-release-binary/action.yml
+++ b/.github/actions/build-release-binary/action.yml
@@ -1,0 +1,50 @@
+name: Build release binary
+description: Builds a release binary
+inputs:
+  target:
+    description: "The Rust target to build for."
+    required: true
+  gh_token:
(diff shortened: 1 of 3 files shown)


klochowicz:
note: I haven't tested it yet as I don't want to get distracted by trying to set up these workflows with different secrets in separate repo.

bonomat (changes requested):
I'm not sure if there is much value in reviewing an untested github action as it will definitely change later on. 

A few notes: 
- given we only have one binary, we don't need this complex action and can go for a predefined [action](https://github.com/marketplace/actions/build-and-push-docker-images)
- there is no dockerfile in the repository
- given we want to run the maker in the cloud and not on an RPi, I don't think we need a multiarch container image

PR #207, opened 2022-11-16
Allow maker to settle a CFD via the UI

### LINKED ISSUE #183: Close CFD
Wiring the frontend/taker with the maker, removing the custom output for closing a cfd (dlc) position.

luckysori:
Fixes #183.

Amounts currently hard-coded.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -262,10 +262,10 @@ pub async fn open_cfd(taker_amount: u64, maker_amount: u64) -> Result<()> {
     tracing::info!("Channels: {binding:?}");
 
     let channel_details = binding.first().context("No first channel found")?;
-    let channel_id = channel_details
+    let maker_pk = channel_details.counterparty.node_id;
+    let short_channel_id = channel_details
         .short_channel_id
         .context("Could not retrieve short channel id")?;
-    let maker_pk = channel_details.counterparty.node_id;
 
     // TODO: Use  MAKER_PK meaningfully
     assert_eq!(maker_pk.to_string(), MAKER_PK, "Using wrong maker seed");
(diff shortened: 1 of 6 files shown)


rust/src/wallet.rs
```diff
@@ -301,7 +301,39 @@ pub async fn open_cfd(taker_amount: u64, maker_amount: u64) -> Result<()> {
...
+    let channels = channel_manager.list_channels();
+    let channel_details = channels.first().context("No channels found")?;
+    let maker_pk = channel_details.counterparty.node_id;
+
+    let taker_amount_msats = taker_amount * 1000;
+    let maker_amount_msats = maker_amount * 1000;
+
+    let maker_connection_str = format!("{maker_pk}@{MAKER_IP_PORT}");
+    let peer_info = parse_peer_info(maker_connection_str)?;
+    connect_peer_if_necessary(&peer_info, peer_manager.clone()).await?;
```
bonomat, line 332: This shouldn't be necessary. Afaik the node will automatically reconnect on restart if there is an open channel. If there is no open channel to the maker, then we shouldn't be in this function anyway :)

bonomat (approved):
Looks good

da-kami (approved):

rust/src/wallet.rs
da-kami: True, we actually found out that it does on my machine and it was a diffferent issue, we could remove this.

luckysori: Done with https://github.com/itchysats/10101/pull/207/commits/b961223edf3a7fba28e1368303f11b1be8319a04.

holzeis (commented):
ups i've never submitted my review. nothing critical, but submitting anyways.

PR #209, opened 2022-11-16
Reusable hardcoded maker connection dets

da-kami:
Probably better to change the API to consume the connection details for now then specifying them hardcoded in rust.

Given that we need to connect to the maker's HTTP api as well this should be in one place - and I think dart should be in control of the state for now.

Note: I did not add the `pubkey` of the maker, but eventually we also have to deal with that. The textfield in `OpenChannel` has to be fixed :)

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -36,6 +36,7 @@ PaymentHistory paymentHistory = PaymentHistory();
 
 const makerIp = "127.0.0.1";
 const makerLightningPort = "9045";
+const makerHttpApiPort = "8000";
 
 void main() {
   FlutterError.onError = (details) {
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -34,6 +34,9 @@ BitcoinBalance bitcoinBalance = BitcoinBalance();
 SeedBackupModel seedBackup = SeedBackupModel();
 PaymentHistory paymentHistory = PaymentHistory();
 
+const makerIp = "127.0.0.1";
+const makerLightningPort = "9045";
+
 void main() {
   FlutterError.onError = (details) {
     FlutterError.presentError(details);
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -80,7 +80,7 @@ pub fn open_channel(peer_pubkey_and_ip_addr: String, channel_amount_sat: u64) ->
 }
 
 #[tokio::main(flavor = "current_thread")]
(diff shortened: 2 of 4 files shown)


rust/src/api.rs
```diff
@@ -80,15 +80,15 @@ pub fn open_channel(peer_pubkey_and_ip_addr: String, channel_amount_sat: u64) ->
 }
 
 #[tokio::main(flavor = "current_thread")]
-pub async fn open_cfd(taker_amount: u64, leverage: u64) -> Result<()> {
+pub async fn open_cfd(taker_amount: u64, leverage: u64, maker_ip_port: String) -> Result<()> {
```
klochowicz, line 83:
```suggestion
pub async fn open_cfd(taker_amount: u64, leverage: u64, maker_http_port: u64 ) -> Result<()> {
```

lib/main.dart
```diff
@@ -34,6 +34,10 @@ BitcoinBalance bitcoinBalance = BitcoinBalance();
 SeedBackupModel seedBackup = SeedBackupModel();
 PaymentHistory paymentHistory = PaymentHistory();
 
+const makerIp = "127.0.0.1";
+const makerLightningPort = "9045";
+const makerHttpApiPort = "8000";
```
klochowicz, line 39: nit: can we specify these as numbers?

klochowicz (approved):
cool!

luckysori:
> Note: I did not add the pubkey of the maker, but eventually we also have to deal with that. The textfield in OpenChannel has to be fixed :)

I suppose we expect the maker's PK (aka node ID) to never change. We might want to feed it in as an environment variable or something similar, so that we can control it depending on the environment (regtest vs testnet).

rust/src/api.rs
klochowicz: sorry, got confused thinking it was only the port!

da-kami:
After a quick discussion with @klochowicz it might be better to just keep the connection details in rust - then we don't have to send them for every call - it's cumbersome to have this in the API... 

We don't actually need this to be configurable in the UI, eventually it *will* be hardcoded to the testnet maker. 
For now it will be best to have a shared regtest seed that we can hardcode the pubkey. We can also allow setting it via an envvar. (same for IP, lightning port, http port). @klochowicz and me are working on that, closing this PR for now.

PR #210, opened 2022-11-16
Hardcode maker details

klochowicz:
Maker pubkey is now hard-coded and matches the checked in seed.

See commit description for further info.


klochowicz:
this PR was an outcome of discussions around #209 - to go as fast as we can, we can just hardcode the counterparty details.

maker/src/main.rs
```diff
@@ -1,14 +1,14 @@
...
-use std::env::temp_dir;
+use std::env::current_dir;
 use std::time::Duration;
 use ten_ten_one::wallet;
 use tracing::metadata::LevelFilter;
 
 #[rocket::main]
 async fn main() -> Result<()> {
-    let path = temp_dir();
+    let path = current_dir()?.join("data").join("maker");
```
bonomat, line 11: much better :) 
Note: we will need to be able to pass this in as an arg once we want to run it in a container.

.gitignore
```diff
@@ -15,3 +15,5 @@ bridge_generated.h
 #ignore keystores
 *.jks
 
+# Artefacts from running maker binary
+/data/maker/
```
bonomat, line 19: 🙃
I love this comment: missing new line

bonomat (approved):
LGTM

maker/src/main.rs
klochowicz: yes, `$DATA_DIR` and `$NETWORK` should soon be added from `clap` :)

da-kami (approved):
Happy to merge this part! 
Once that is in we should also change `open_channel` API to use these values rather than passing the connection string from the user interface.
We will then need an API endpoint to request the "current" maker from the rust code.

.gitignore
klochowicz: added :)

klochowicz:
> Happy to merge this part!
> Once that is in we should also change `open_channel` API to use these values rather than passing the connection string from the user interface.
> We will then need an API endpoint to request the "current" maker from the rust code.

great idea! I'll continue working on that straight away.

klochowicz:
> Happy to merge this part!
> Once that is in we should also change `open_channel` API to use these values rather than passing the connection string from the user interface.
> We will then need an API endpoint to request the "current" maker from the rust code.

addressed in #214

Issue #201, opened 2022-11-15
Persist CFD data and custom output data for display and collab close on the taker


da-kami:
The taker has to remember the `CustomOutputDetails` received during opening the CFD.

We will need a database to store this information alongside with the data that captured the opening ("initial") values of the Cfd.
You can refer to the `CfdDetail` screen and ItchySats to see what we need stored.

bonomat:
https://github.com/itchysats/10101/pull/211 loads the payment history from lighting and displays it in the ui. Within this ticket I would expect to extend the list of `LightningTransactions` to also include the `Cfds` transactions.

da-kami:
After discussion with @luckysori for our current solution with exactly one channel and one CFD we actualy don't have to store the `CustomOutputDetail` because we can just get it with `channel.first().outputs..first()` from the channel manager.

However, we will have to sage the initial values of the CFD to be able to come up with the payouts according to the price change. Note: This will require some kind of payout curve calc.

bonomat:
> This will require some kind of payout curve calc as described in https://github.com/itchysats/10101/issues/219

No need for a payout curve, a simple formula like
`pnl = quantity_usd* ((1/price_opening)-(1/price_closing)`
should be sufficient.

da-kami:
I'm closing this one with #231 - for any needed follow-ups please create specific tickets!

PR #211, opened 2022-11-16
Lightning payment history

### LINKED ISSUE #165: Payment history

bonomat:
Resolves #165 

Note: 
* What is missing is the overview of the actual CFDs. I think we don't have this data yet.
* We cannot know easily which transaction opened the channel. For that we will need to store some additional data somewhere.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -1,4 +1,5 @@
 use crate::lightning;
+use crate::lightning::HTLCStatus;
 use crate::lightning::LightningSystem;
 use crate::lightning::PeerInfo;
 use crate::seed::Bip39Seed;
@@ -11,6 +12,7 @@ use anyhow::Result;
 use bdk::bitcoin;
 use bdk::bitcoin::secp256k1::PublicKey;
 use bdk::bitcoin::Address;
+use bdk::bitcoin::Amount;
 use bdk::bitcoin::Script;
 use bdk::bitcoin::Txid;
 use bdk::blockchain::ElectrumBlockchain;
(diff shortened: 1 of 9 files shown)


klochowicz (approved):
LGTM!

da-kami:
> Resolves #165
> 
> Note:
> 
>     * What is missing is the overview of the actual CFDs. I think we don't have this data yet.

This is tracked in https://github.com/itchysats/10101/issues/201
It will be enabled by: https://github.com/itchysats/10101/pull/220

>     * We cannot know easily which transaction opened the channel. For that we will need to store some additional data somewhere.

Yeah, we would have to store metadata in the database for that as well! I added https://github.com/itchysats/10101/issues/226 for this task - feel free to extend it if the description is not complete! 
It does not have the highest prio for me, but I think we can easily add this feature once we have the database (https://github.com/itchysats/10101/pull/220)

rust/src/wallet.rs
```diff
@@ -228,6 +230,41 @@ pub fn get_bitcoin_tx_history() -> Result<Vec<bdk::TransactionDetails>> {
...
+            status: payment_info.status.clone().into(),
+            timestamp: payment_info.timestamp,
+        })
+        .collect();
+
+    let mut inbound = inbound
+        .iter()
+        .map(|(_, payment_info)| LightningTransaction {
+            tx_type: LightningTransactionType::Payment,
+            flow: Flow::Outbound,
```
da-kami, line 257:
```suggestion
            flow: Flow::Inbound,
```

🛠️ I think this is a copy-paster that should be fixed :D

da-kami (approved):
Noice, but I think there is a copy-paster that needs a fix, see comment :)

A screenshot would be great :)

rust/src/wallet.rs
bonomat: Good catch. Fixed

bonomat:
> A screenshot would be great :)

Didn't change anything in the layout, so the old mocks would have shown the same :) 

bors r+

da-kami:
> > A screenshot would be great :)
> 
> Didn't change anything in the layout, so the old mocks would have shown the same :)
> 
> bors r+

Sure, but it's real data now and not mocked garbage :D

bonomat:
> Sure, but it's real data now and not mocked garbage :D

you wouldn't know the difference because the mocked data was such high quality and accurately 😅

PR #214, opened 2022-11-16
chore: Do not specify maker peer info manually

klochowicz:
Pass in the hard-coded values to flutter. Expose them in the API so that they
     can be displayed.

### DIFF
--- a/lib/wallet/open_channel.dart
+++ b/lib/wallet/open_channel.dart
@@ -26,6 +26,7 @@ class _OpenChannelState extends State<OpenChannel> {
     super.initState();
     final bitcoinBalance = context.read<BitcoinBalance>();
     channelCapacity = bitcoinBalance.amount.asSats;
+    _setMakerPeerInfo();
   }
 
   @override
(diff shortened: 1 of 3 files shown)


holzeis:
I don't think thats the right approach. We should pass the configuration using the cli. see also https://github.com/itchysats/10101/pull/215

holzeis (changes requested):
I think we can skip that PR in favour of https://github.com/itchysats/10101/pull/215

klochowicz:
> I think we can skip that PR in favour of #215

is there anything wrong with this PR? you've requested changes. If your one goes further, then it always can be rebased on top.

da-kami:
> I think we can skip that PR in favour of #215

I think it is totally fine to do this in rust, and doing it in flutter makes our life harder because we have to communicate it to rust.

da-kami (approved):

holzeis:
> > I think we can skip that PR in favour of #215
> 
> I think it is totally fine to do this in rust, and doing it in flutter makes our life harder because we have to communicate it to rust.

I prefer a configurable app over setting hardcoded configs.

PR #215, opened 2022-11-16
feat: parse flutter arguments

holzeis:
This will allow for providing configurations using the command line when running flutter.

e.g.
```
flutter run -d macos --dart-define="NETWORK=testnet" --dart-define="MAKER_P2P_ENDPOINT=pubkey@127.0.0.1:9045"
```

The following parameter are supported
- MAKER_IP
- MAKER_PK
- MAKER_PORT_LIGHTNING
- MAKER_PORT_HTTP
- MAKER_P2P_ENDPOINT (will overwrite MAKER_IP, MAKER_PK and MAKER_PORT_LIGHTNING if set)

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -13,6 +13,7 @@ import 'package:ten_ten_one/cfd_trading/cfd_order_detail.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/models/payment.model.dart';
 import 'package:ten_ten_one/payment_history_change_notifier.dart';
+import 'package:ten_ten_one/utilities/environment.dart';
 import 'package:ten_ten_one/wallet/deposit.dart';
 import 'package:ten_ten_one/wallet/open_channel.dart';
 import 'package:ten_ten_one/wallet/wallet.dart';
@@ -41,10 +42,9 @@ void main() {
     FlutterError.presentError(details);
   };
 
-  final config = FLog.getDefaultConfigurations();
-  config.activeLogLevel = foundation.kReleaseMode ? LogLevel.INFO : LogLevel.DEBUG;
-
-  FLog.applyConfigurations(config);
+  final logConfig = FLog.getDefaultConfigurations();
+  logConfig.activeLogLevel = foundation.kReleaseMode ? LogLevel.INFO : LogLevel.DEBUG;
+  FLog.applyConfigurations(logConfig);
 
   runApp(MultiProvider(providers: [
     ChangeNotifierProvider(create: (context) => lightningBalance),
@@ -54,6 +54,7 @@ void main() {
     ChangeNotifierProvider(create: (context) => CfdTradingChangeNotifier()),
     ChangeNotifierProvider(create: (context) => WalletChangeNotifier()),
     ChangeNotifierProvider(create: (context) => cfdOffersChangeNotifier),
+    Provider(create: (context) => Environment.parse()),
   ], child: const TenTenOneApp()));
 }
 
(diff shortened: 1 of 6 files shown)


klochowicz:
would it not be easier to just set the env variables and read them directly in Rust? `--dart-define` seems like a hack.

we're in fact already directly parsing some env variables, see `RUST_LOG` handling.

klochowicz (commented):
I think it's faster to go directly into Rust, there's little point in parsing stuff in flutter beforehand if it's not used at all.

da-kami:
What is the motivation of this? I feel it makes our life unnecessarily hard.

holzeis:
> would it not be easier to just set the env variables and read them directly in Rust? `--dart-define` seems like a hack.
> 
> we're in fact already directly parsing some env variables, see `RUST_LOG` handling.

If you only want to use these variables for development you are right, but if you want to use them natively `--dart-define`is the way to go.

holzeis:
> I think it's faster to go directly into Rust, there's little point in parsing stuff in flutter beforehand if it's not used at all.

we need to go that path if we want to have a configurable app.

holzeis:
> What is the motivation of this? I feel it makes our life unnecessarily hard because we don't ever use these connection strings in dart - we only use it in rust so we might as well set it there :)

like mentioned above, if you want to be able to configure your flutter app, without building a new release. That's the way to go.

holzeis:
We might be able to access the environment variables in the rust code though, if you dislike passing the configuration via an api call.

holzeis:
> We might be able to access the environment variables in the rust code though, if you dislike passing the configuration via an api call.

Just tested it. It's unfortunately not possible to access the environment variable (set through --dart-define) from rust.

klochowicz:
> Just tested it. It's unfortunately not possible to access the environment variable (set through --dart-define) from rust.

you don't need `--dart-define` - it's unnecessary. You can use env variables just like in any other unix program.
you can actually see this working when you call `RUST_LOG=bdk:debug flutter run -d macos` (`RUST_LOG` parsing was done ages ago, that's why I pointed you to it)

holzeis:
> > Just tested it. It's unfortunately not possible to access the environment variable (set through --dart-define) from rust.
> 
> you don't need `--dart-define` - it's unnecessary. You can use env variables just like in any other unix program. you can actually see this working when you call `RUST_LOG=bdk:debug flutter run -d macos` (`RUST_LOG` parsing was done ages ago, that's why I pointed you to it)

Ok, assume you want to set the RUST_LOG variable on a released app (not for development). The only way to set this environment variable is through `--dart-define="RUST_LOG=...`. However, this environment variable is unfortunately not accessible directly from rust (just tested) - but only from your dart code.

klochowicz:
> > > Just tested it. It's unfortunately not possible to access the environment variable (set through --dart-define) from rust.
> > 
> > 
> > you don't need `--dart-define` - it's unnecessary. You can use env variables just like in any other unix program. you can actually see this working when you call `RUST_LOG=bdk:debug flutter run -d macos` (`RUST_LOG` parsing was done ages ago, that's why I pointed you to it)
> 
> Ok, assume you want to set the RUST_LOG variable on a released app (not for development). The only way to set this environment variable is through `--dart-define="RUST_LOG=...`. However, this environment variable is unfortunately not accessible directly from rust (just tested) - but only from your dart code.

sorry, I think we're talking across purposes. There's currently no need to change anything for *released* app, I'd imagine that one should have it's config hard-coded for testnet. Let's remember that there's no command line shell on an Android phone.

holzeis:
> 

fair enough. I might have tunneled myself into a solution that seems to bring no additional value.

klochowicz:
> > 
> 
> fair enough. I might have tunneled myself into a solution that seems to bring no additional value.

all good, this happens to all of us sometimes 🙂

PR #216, opened 2022-11-16
docs: fifth lol ppp

holzeis:
already published, sicne we are already late. But happy to update in case you have any feedback / objections.

https://makers.bolt.fun/story/ppps-week-5-10101--397

### DIFF
--- a/docs/ppp/iteration-5-16112022.md
+++ b/docs/ppp/iteration-5-16112022.md
@@ -1,0 +1,32 @@
+### Plans 📆
+
+- Create invoices
+- Send Payments
+- Bug fixing and UX improvements
+- Replacing mock data with real data
+- Show payment history
+- Containerize maker
+- Wiring up close (settle) CFD
+- Persisting CFDs
+
+### Progress ✅
+
+- Fund wallet screen
+- Setup maker service with ldk integration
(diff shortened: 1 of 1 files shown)


docs/ppp/iteration-5-16112022.md
```diff
@@ -0,0 +1,32 @@
...
+- Wired up open CFD (still with mock amounts)
+- Improved logging
+- Started ldk node on the phone
+- Open channel with maker
+- Fetch bid, ask and index prices from bitmex
+- Show history of on-chain transactions
+
+### Problems ✋
+
+- None, after a busy week, we are on good track. Still a lot to be done, but only doing from here on.
```
bonomat, line 26: Shall we be honest about our short cuts and the outcome of our LDK extension? 

- When implementing our custom protocol on the LDK which allows us to add DLCs to the channel we accepted one short coming: while a DLC is added, you cannot accept or send a payment. This would break the DLC and would make the outputs unspendable. For the tournament we decided to accept this shortcoming but we will focus on this afterwards.

bonomat (approved):
Thanks for the writeup.

docs/ppp/iteration-5-16112022.md
holzeis: I am going to add that.

PR #218, opened 2022-11-16
feat: allow 0-conf channels

### LINKED ISSUE #206: Make a newly added channel usable after one confirmation
When we open a channel we currently have to wait for `7` blocks until the channel is actually usable - on `testnet` this can take significant time. 

The user needs an indicator when this target is reached and trading is possible, which will be covered (to some extent) in #199. But we cannot let the user hanging for 7 confirmations, so we should find out how to configure taker and maker to reduce the confirmations to a minimum.

bonomat:
While not recommended in production, we allow 0-conf channels. To do so we need to manually confirm all channel requests.

resolves #206

### DIFF
--- a/rust/src/lightning.rs
+++ b/rust/src/lightning.rs
@@ -140,6 +140,8 @@ pub async fn open_channel(
         },
         channel_handshake_config: ChannelHandshakeConfig {
             announced_channel: false,
+            // We want to support 0-conf channels.
+            minimum_depth: 0,
             ..Default::default()
         },
         ..Default::default()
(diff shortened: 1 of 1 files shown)


rust/src/lightning.rs
```diff
@@ -331,6 +333,11 @@ pub fn setup(
     user_config
         .channel_handshake_limits
         .force_announced_channel_preference = false;
+    // By setting this to 0 we indicate we accept 0-conf channels.
+    user_config.channel_handshake_config.minimum_depth = 0;
+    // By setting `manually_accept_inbound_channels` to `true` we need to manually confirm every
+    // inbound channel request.
+    user_config.manually_accept_inbound_channels = true;
```
luckysori, line 340: ❓ Is this actually necessary?

Below you're just accepting unconditionally anyway.

luckysori (approved):

klochowicz (approved):

rust/src/lightning.rs
bonomat: Yes, otherwise the event `OpenChannelRequest` won't be fired

PR #220, opened 2022-11-17
feat: Add persistent SQLite database

klochowicz:
I've used sqlx purely as we are familiar with it already. Otherwise, it's a bit of a pain
to use an async database here - a sync one would be much easier (due to frb
async boilerplate and runtime cost).

Database is managed for the lifetime of the app by the `ten_ten_one` library,
after being initialised with `db::init_db()` call.

Database in empty for now, but initialising and acquiring a connection works for
both maker and taker.

Note: There's no need to manually close connections to the DB, because we are
only storing a ConnectionPool, and the connections are closed with RAII when
they get out scope.

### DIFF
--- a/rust/src/db.rs
+++ b/rust/src/db.rs
@@ -1,0 +1,47 @@
+use anyhow::anyhow;
+use anyhow::Context;
+use anyhow::Result;
+use sqlx::pool::PoolConnection;
+use sqlx::sqlite::SqliteConnectOptions;
+use sqlx::Sqlite;
+use sqlx::SqlitePool;
+use state::Storage;
+use std::path::Path;
+use std::sync::Mutex;
+use std::sync::MutexGuard;
+
+pub type SqliteConnection = PoolConnection<Sqlite>;
+
+/// Wallet has to be managed by Rust as generics are not support by frb
+static DB: Storage<Mutex<SqlitePool>> = Storage::new();
+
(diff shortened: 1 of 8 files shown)


da-kami (approved):

rust/src/wallet.rs
```diff
@@ -55,6 +55,7 @@ pub fn maker_peer_info() -> PeerInfo {
     }
 }
 
+#[derive(Clone, derive_more::Display)]
```
bonomat, line 58: Not necessary to action, but for the future: I don't think we should add an extra library for this but just implement it ourselves.

bonomat (approved):
Looks good.

rust/src/db.rs
```diff
@@ -0,0 +1,47 @@
...
+use sqlx::Sqlite;
+use sqlx::SqlitePool;
+use state::Storage;
+use std::path::Path;
+use std::sync::Mutex;
+use std::sync::MutexGuard;
+
+pub type SqliteConnection = PoolConnection<Sqlite>;
+
+/// Wallet has to be managed by Rust as generics are not support by frb
```
holzeis, line 15: nit: copy & paste error

Wallet -> SqlitePool

holzeis (approved):

rust/src/db.rs
klochowicz: thanks, I'll remove the type name from the comment in both cases, it's irrelevant!

PR #222, opened 2022-11-17
Avoid unhandled exceptions during frontend syncs

luckysori:
The main problem that we are fixing is the fact that the possible exception derived from calling `api.getOffer` was unhandled.

Since we already handle exceptions by default when calling `runPeriodically`, by removing the initial syncs (for the offer,
wallet and payment history) we ensure that all those calls cannot kill the application by running into an unhandled exception.

In order to still have an initial sync, we modify `runPeriodically` to immediately call the `callback` argument and only afterwards set up the periodic timer.


klochowicz (approved):
thanks for solving this 👍

PR #224, opened 2022-11-17
Implement functionality to send bitcoin on-chain

### LINKED ISSUE #114: Withdraw funds from on-chain wallet
Like the title suggest, the possibility to withdraw funds from the wallet.

Image

luckysori:
Fixes #114.

https://user-images.githubusercontent.com/9418575/202347714-cb78e967-8b16-46dd-a772-fff67a1a5aee.mp4

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -13,7 +13,7 @@ import 'package:ten_ten_one/cfd_trading/cfd_order_detail.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/models/payment.model.dart';
 import 'package:ten_ten_one/payment_history_change_notifier.dart';
-import 'package:ten_ten_one/wallet/deposit.dart';
+import 'package:ten_ten_one/wallet/receive_on_chain.dart';
 import 'package:ten_ten_one/wallet/open_channel.dart';
 import 'package:ten_ten_one/wallet/wallet.dart';
 import 'package:ten_ten_one/wallet/wallet_change_notifier.dart';
@@ -113,9 +113,9 @@ class _TenTenOneState extends State<TenTenOneApp> {
               },
             ),
             GoRoute(
-              path: Deposit.subRouteName,
+              path: ReceiveOnChain.subRouteName,
               builder: (BuildContext context, GoRouterState state) {
-                return const Deposit();
+                return const ReceiveOnChain();
               },
             ),
             GoRoute(
@@ -232,7 +232,7 @@ class _TenTenOneState extends State<TenTenOneApp> {
             bitcoinTxHistoryItem.sent != 0
                 ? Amount(bitcoinTxHistoryItem.sent * -1)
                 : Amount(bitcoinTxHistoryItem.received),
-            bitcoinTxHistoryItem.sent != 0 ? PaymentType.sendOnChain : PaymentType.deposit,
+            bitcoinTxHistoryItem.sent != 0 ? PaymentType.sendOnChain : PaymentType.receiveOnChain,
             bitcoinTxHistoryItem.isConfirmed ? PaymentStatus.finalized : PaymentStatus.pending,
             bitcoinTxHistoryItem.timestamp))
         .toList();
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -26,7 +26,7 @@ import 'package:ten_ten_one/wallet/receive.dart';
 import 'package:ten_ten_one/wallet/seed.dart';
 import 'package:go_router/go_router.dart';
 import 'package:ten_ten_one/wallet/send.dart';
-import 'package:ten_ten_one/wallet/withdraw.dart';
+import 'package:ten_ten_one/wallet/send_on_chain.dart';
 import 'package:ten_ten_one/bridge_generated/bridge_definitions.dart' as bride_definitions;
 
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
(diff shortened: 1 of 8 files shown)


lib/models/payment.model.dart
```diff
@@ -32,7 +32,7 @@ enum PaymentType {
 extension PaymentTypeExtension on PaymentType {
   static const displays = {
     PaymentType.deposit: "Bitcoin Deposited",
-    PaymentType.withdraw: "Bitcoin Withdrawn",
+    PaymentType.sendOnChain: "Bitcoin Sent",
```
klochowicz, line 35: IMHO `Withdrawn` was better, especially when he had the counterparty "Deposited".
If you really want to go for "Send", the other action should rather be "Receive"

rust/src/wallet.rs
```diff
@@ -156,6 +161,43 @@ impl Wallet {
         Ok(address.address)
     }
 
+    pub fn send_to_address(&self, address: Address, amount: u64) -> Result<Txid> {
```
da-kami, line 164: I implemented basically the same in https://github.com/itchysats/10101/pull/225 but I am estimating the fee using electrum client (hope that works correctly).
We will have to align this a bit I guess :)

rust/src/wallet.rs
```diff
@@ -156,6 +161,43 @@ impl Wallet {
...
+        let electrum_url = match self.lightning.network {
+            bitcoin::Network::Bitcoin => MAINNET_ELECTRUM,
+            bitcoin::Network::Testnet => TESTNET_ELECTRUM,
+            bitcoin::Network::Regtest => REGTEST_ELECTRUM,
+            _ => unreachable!("Signet is not supported"),
+        };
+
+        let client = Client::new(electrum_url)?;
+
+        client.transaction_broadcast(&tx)?;
```
klochowicz, line 196: that functionality is already exposed in the `BdkLdkWallet` (it has a `Client` inside)

lib/models/payment.model.dart
luckysori: But surely when you are paying someone on-chain with your wallet you are not withdrawing coins, but rather sending them.

rust/src/wallet.rs
luckysori: Nice! Although @da-kami left a TODO that makes me think using that one isn't perfect?

``` rust
// TODO: This swallows the result internally - if we fail to broadcast we'll never know :/
self.lightning.wallet.broadcast_transaction(&tx);
```

lib/wallet/send_on_chain.dart
```diff
@@ -11,18 +17,83 @@ class SendOnChain extends StatefulWidget {
...
+                        children: [
+                          Row(
+                            mainAxisAlignment: MainAxisAlignment.end,
+                            children: [
+                              ElevatedButton(
+                                  onPressed: () async {
+                                    FLog.info(
+                                        text:
+                                            "Sending " + amount.toString() + " sats to " + address);
+                                    await api.sendToAddress(address: address, amount: amount);
```
holzeis, line 84: 🤔 Can this call fail? and if yes, shouldn't we add an error handling? at least showing the user something did not work?

holzeis (commented):
the first 3 commits look like they don't belong to this PR.

da-kami:
I'll review again once rebased :)

lib/models/payment.model.dart
klochowicz: 1. most of the lightning wallets seem to limit the btc wallet functionality - up to a point of completely hiding it in some cases - it's only used to deposit/withdraw money, because send/receive should rather be made via lightning.
2. it's called 'withdraw' in ItchySats too, and also allows moving funds to any other arbitrary on-chain wallet.

what do you think about `Send/Receive`? (just like for lightning). one additional side-effect of the old naming was also the clear distinction between bitcoin/lightning actions.

rust/src/wallet.rs
klochowicz: #229 solves this :)

lib/models/payment.model.dart
bonomat: I'm not convinced anymore that 1) is a good idea. I'd like a wallet which can speak both, so that I can use funds on-chain if needed.

at 2) `withdraw` on ItchySats made sense because it was not designed to be a wallet. Here it is a wallet. Imho it should be `send/receive` here as well

Adding an additional point :)

3) I'm not a big fan of those labels in general, I think this can be designed better.

klochowicz: `send/receive` works for me too! what irked me a bit was the `deposit/send` disjoint.

labels/icons can be easily changed, but the vocabulary of the UI matters :)

bonomat: I like this design: 
`You send, on-chain` vs `You sent, Lightning`

image

klochowicz: > I'm not convinced anymore that 1) is a good idea. I'd like a wallet which can speak both, so that I can use funds on-chain if needed.

this might as well be something that differentiates us a bit from other wallets; I like the explicitness of it and the additional useful feature.

lib/main.dart
```diff
@@ -209,7 +202,7 @@ class _TenTenOneState extends State<TenTenOneApp> {
             bitcoinTxHistoryItem.sent != 0
                 ? Amount(bitcoinTxHistoryItem.sent * -1)
                 : Amount(bitcoinTxHistoryItem.received),
-            bitcoinTxHistoryItem.sent != 0 ? PaymentType.withdraw : PaymentType.deposit,
+            bitcoinTxHistoryItem.sent != 0 ? PaymentType.send : PaymentType.deposit,
```
da-kami, line 205:
```suggestion
            bitcoinTxHistoryItem.sent != 0 ? PaymentType.sendOnChain : PaymentType.deposit,
```

da-kami, line 205:
```suggestion
            bitcoinTxHistoryItem.sent != 0 ? PaymentType.send : PaymentType.receiveOnChain,
```

rust/src/wallet.rs
luckysori: Thank you, this is dealt with now!

da-kami (approved):

PR #225, opened 2022-11-17
Maker to send funds on open channel

### LINKED ISSUE #199: Send funds from maker to taker for channel opening
In our regtest setup the taker currently has funds and just opens the channel sending 2/3 of the channel capacity to the maker. 

This does not not reflect the flow we agreed upon for the tournament:

1. Contact the maker for opening a channel (HTTP endpoint)
    1. The taker provides a Bitcoin address for funding
    2. The taker defines "how much funds the taker" wants to use for trading in the channel; the channel capacity is 1/3 taker, 2/3 maker.
     1.  The taker's `Open Channel` screen should not talk about the capacity, that's confusing - it should talk about how much funds the taker wants to use for trading in the channel.
     1. (The maker could respond with maker's connection details) -> Hardcode the maker details in the taker will do
1. The maker sends the 2/3 of funds to the taker's Bitcoin address
    1. I think the taker's user interface should not  show this transaction otherwise that might be confusing, we might want to filter this address when displaying the history.
1. The taker opens a channel with the taker+maker amount and sends the maker BTC back to the maker as part of channel opening
    1. The taker can only open the channel once the funds of the maker are in mempool, so we will need an `Opening Channel` state in the payment list which spins until the channel is open. This is spinning while:
        1. Send the message to the maker
        2. Maker sends BTC 
        3. Taker sees BTC in mempool
        4. Taker triggers opening channel.  
        5. The channel is open - and usable...? see: https://github.com/itchysats/10101/issues/206

Notes:

- I don't think we have to keep any state on the maker side; the maker just sends out BTC when a taker's request comes in and expects that a channel will be opened and funds will be sent over through the channel open. (the maker acts as a testnet faucet :)
- The taker might have to persist the address that is used for channel opening so we can filter that address and not display it in the history (to avoid confusion :)

---

Note: This is obviously not safe for the maker and simplified to get a working version for the tournament. This should *never* be used in production!

da-kami:
resolves #199

The part where we "ignore" the makers tx for funding (on bitcoin history and balance) is not implemented yet. 
I am planning to work on this in a follow-up PR on top because it is a  bit tricky to get right.

Note: Still testing if it works :)

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -49,7 +49,7 @@ pub async fn get_offer(
     })
 }
 
-#[rocket::post("/channel/<remote_node_id>")]
+#[rocket::post("/channel/close/<remote_node_id>")]
 pub async fn post_force_close_channel(remote_node_id: String) -> Result<(), HttpApiProblem> {
     let remote_node_id = remote_node_id.parse().map_err(|e| {
         HttpApiProblem::new(StatusCode::BAD_REQUEST)
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -9,6 +9,9 @@ use rocket::State;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use ten_ten_one::wallet::force_close_channel;
+use ten_ten_one::wallet::send_to_address;
+use ten_ten_one::wallet::OpenChannelRequest;
+use ten_ten_one::wallet::OpenChannelResponse;
 use tokio::sync::watch;
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -61,4 +64,18 @@ pub async fn post_force_close_channel(remote_node_id: String) -> Result<(), Http
     })?;
 
     Ok(())
+}
+
+#[rocket::post("/channel/open", data = "<request>", format = "json")]
+pub async fn post_open_channel(
+    request: Json<OpenChannelRequest>,
+) -> Result<Json<OpenChannelResponse>, HttpApiProblem> {
+    let funding_txid = send_to_address(request.address_to_fund.clone(), request.fund_amount)
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Failed to open channel with maker")
+                .detail(format!("Failed to transfer funds: {e:#}"))
+        })?;
+
+    Ok(Json(OpenChannelResponse { funding_txid }))
 }
--- a/rust/src/lightning.rs
+++ b/rust/src/lightning.rs
@@ -77,8 +77,6 @@ use std::time::Duration;
 use std::time::SystemTime;
 use tokio::task::JoinHandle;
 
-pub use lightning::*;
-
 /// Container to keep all the components of the lightning network in one place
 #[derive(Clone)]
 pub struct LightningSystem {
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -3,20 +3,28 @@ use crate::lightning::LightningSystem;
 use crate::lightning::PeerInfo;
 use crate::seed::Bip39Seed;
 use anyhow::anyhow;
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use bdk::bitcoin;
 use bdk::bitcoin::secp256k1::PublicKey;
+use bdk::bitcoin::Script;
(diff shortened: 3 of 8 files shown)


maker/src/routes.rs
```diff
@@ -62,3 +65,17 @@ pub async fn post_force_close_channel(remote_node_id: String) -> Result<(), Http
 
     Ok(())
 }
+
+#[rocket::post("/open_channel", data = "<request>")]
```
klochowicz, line 69: nit: it would be nice if we could align the API of opening and closing of the channel, e.g.
- `/api/channel/close/<id>`
- `/api/channel/open/<body>`

rust/src/lightning.rs
```diff
@@ -77,8 +77,6 @@ use std::time::Duration;
 use std::time::SystemTime;
 use tokio::task::JoinHandle;
 
-pub use lightning::*;
-
```
klochowicz, line 81: <3

rust/src/mempool.rs
```diff
@@ -0,0 +1,51 @@
+use crate::cli::Network;
```
klochowicz: remove the unused module :-)

rust/src/wallet.rs
```diff
@@ -342,6 +455,12 @@ pub async fn force_close_channel(remote_node_id: PublicKey) -> Result<()> {
     Ok(())
 }
 
+pub fn transfer(address: Address, amount: u64) -> Result<Txid> {
+    let txid = get_wallet()?.transfer(address, amount)?;
+
+    Ok(txid)
```
klochowicz, line 461: 🙃
```suggestion
     get_wallet()?.transfer(address, amount)
```

da-kami:
@bonomat I did fee estimation using electrum - in ItchySats we used mempool for it - any specific reason why we did not use electrum?

klochowicz (approved):

rust/src/wallet.rs
```diff
@@ -182,6 +195,48 @@ impl Wallet {
...
+        let script_status = self
+            .lightning
+            .wallet
+            .get_tx_status_for_script(script, txid)
+            .map_err(|_| anyhow!("Could not get tx status for script"))?;
+        Ok(script_status)
+    }
+
+    pub fn send_to_address(&self, send_to: Address, amount: u64) -> Result<Txid> {
+        let wallet = self.lightning.wallet.get_wallet();
```
luckysori, line 209: ❓ Shouldn't we sync with the blockchain like we do in ItchySats?

rust/src/wallet.rs
```diff
@@ -182,6 +195,48 @@ impl Wallet {
...
+
+        let estimated_fee_rate = self
+            .lightning
+            .wallet
+            .estimate_fee(ConfirmationTarget::Normal)
+            .map_err(|_| anyhow!("Failed to estimate fee"))?;
```
luckysori, line 215: 👍

rust/src/wallet.rs
```diff
@@ -182,6 +195,48 @@ impl Wallet {
...
+        };
+
+        if !wallet.sign(&mut psbt, SignOptions::default())? {
+            bail!("Failed to sign psbt");
+        }
+
+        let tx = psbt.extract_tx();
+
+        // TODO: This swallows the result internally - if we fail to broadcast we'll never know :/
+        self.lightning.wallet.broadcast_transaction(&tx);
```
luckysori, line 236: 👍 Nice that this exists. Can we fix the problem in the TODO easily? 

The alternative is to build an Electrum client here. Calling `transaction_broadcast` on it does return a `Result`.

luckysori (commented):
Only looked at the overlap between this PR and https://github.com/itchysats/10101/pull/224.

It looks good in that sense. Let's merge this first and I'll adapt accordingly!

rust/src/wallet.rs
da-kami: I don't remember why this was needed, but I'm happy to add it!

rust/src/wallet.rs
klochowicz: I'm just fixing this TODO with a PR against `bdk-ldk`

klochowicz (approved):
thanks for incorporating previous feedback and making it work ;-)

bonomat:
> @bonomat I did fee estimation using electrum - in ItchySats we used mempool for it - any specific reason why we did not use electrum?

Because in maker automation we did not have electrum available.

Issue #200, opened 2022-11-15
Basic value validation / value & action restriction for the CFD order


da-kami:
- [x] Make sure the margin is within channel bounds (i .e. the taker can only trade 1/3 of the channel capacity
- [x] Make sure the margin is within balance bounds (the margin cannot be higher than balance)

- [x] Make sure that we cannot trigger opening a CFD if we don't have a channel / channel balance

Notes: 

- Restricting the leverage is out of scope, captured in #198
- The margin is hard-coded after https://github.com/itchysats/10101/pull/196 is merged (see https://github.com/itchysats/10101/pull/196#discussion_r1023374713).

PR #227, opened 2022-11-17
fix: only allow btc usd as traiding pair

holzeis:
We currently don't support ETHUSD, hence removing the option for now,

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -95,7 +95,7 @@ class _CfdOfferState extends State<CfdOffer> {
             value: order.quantity.toString(),
           ),
           Dropdown(
-              values: TradingPair.values.map((t) => t.name.toUpperCase()).toList(),
+              values: [TradingPair.btcusd.name.toUpperCase()],
               onChange: (tradingPair) {
                 order.tradingPair =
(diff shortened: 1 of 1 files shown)


lib/cfd_trading/cfd_offer.dart
```diff
@@ -95,7 +95,9 @@ class _CfdOfferState extends State<CfdOffer> {
             value: order.quantity.toString(),
           ),
           Dropdown(
-              values: TradingPair.values.map((t) => t.name.toUpperCase()).toList(),
+              // Only BTCUSD is currently supported, hence I am removing the list here for now.
```
da-kami, line 98:
```suggestion
```

lib/cfd_trading/cfd_offer.dart
```diff
@@ -95,7 +95,9 @@ class _CfdOfferState extends State<CfdOffer> {
...
+              // TradingPair.values.map((t) => t.name.toUpperCase()).toList(),
```
da-kami, line 99:
```suggestion
```

da-kami (approved):
Happy to merge this. I would remove the comments.

Note: We could instead add "coming soon" messages when we implement the validation in #200 - but probably easier to just remove it, similar to restricting the leverage

klochowicz (approved):

PR #230, opened 2022-11-17
chore: Do not run Android integration tests on bors

klochowicz:
We're not using this job when deriving bors status, so we can avoid running it.
Hopefully this should make other job runners wait shorter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
   # See: https://betterprogramming.pub/test-flutter-apps-on-android-with-github-actions-abdba2137b4
   flutter-android-test:
     needs: changes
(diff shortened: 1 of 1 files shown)


klochowicz:
as you can tell, we're skipping integration tests here for now. :)

holzeis (approved):
@klochowicz I think this makes sense, as we don't maintain our integration tests anyways. At this point I guess a failing tests is more likely to be an issue with the test rather with the code.

klochowicz:
> @klochowicz I think this makes sense, as we don't maintain our integration tests anyways. At this point I guess a failing tests is more likely to be an issue with the test rather with the code.

failing test might indicate rather than we added a feature / dependency that does not work on Android. this is my only worry at this stage - but with running it still on `main` we still have a chance to see that.

PR #231, opened 2022-11-17
Introduce CFD model and persist Cfd data

da-kami:
The Cfd model is to be filled with data from the rust backend. We periodically update the list of Cfds by calling the backend which loads the data from the db. We insert/update/delete the Cfd in the database according to user actions (e.g. open cfd inserts a cfd, ...).

We keep an `Order` model that is used before the CFD exists (i.e. before the custom output is added). The Order is not persisted for the duration of the tournament. We don't deal with failed `Order`s yet, but display them as failed Cfds. This is similar to what we did in ItchySats.

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -1,5 +1,3 @@
-import 'dart:math';
-
 import 'package:flutter/material.dart' hide Divider;
 import 'package:go_router/go_router.dart';
 import 'package:intl/intl.dart';
@@ -12,10 +10,10 @@ import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
 import 'package:ten_ten_one/cfd_trading/position_selection.dart';
 import 'package:ten_ten_one/models/amount.model.dart';
-import 'package:ten_ten_one/models/order.dart';
 import 'package:ten_ten_one/utilities/divider.dart';
 import 'package:ten_ten_one/utilities/dropdown.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
+import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 
 class CfdOffer extends StatefulWidget {
   static const leverages = [1, 2];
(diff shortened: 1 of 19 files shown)


rust/prepare_db.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
holzeis, line 1: ❓ I've seen that in itchysats, but what is this used for? the migration will be run automatically in the code.

rust/src/cfd.rs
```diff
@@ -141,6 +142,8 @@ pub async fn open(order: &Order) -> Result<()> {
 
     let mut connection = db::acquire().await?;
 
+    // TODO: Remove these weird wrapper types, should not be needed for what we are doing
```
holzeis: 👍 is this possible? I tried without the wrapper types and cargo was complaining that e.g. i32 is not implementing the sqlx type trait. That's the reason I've added the weird wrapper types 😅.

rust/src/db.rs
```diff
@@ -50,3 +52,50 @@ fn get_db() -> Result<MutexGuard<'static, SqlitePool>> {
         .lock()
         .map_err(|_| anyhow!("cannot acquire DB lock"))
 }
+
+pub async fn load_cfds(conn: &mut SqliteConnection) -> Result<Vec<Cfd>> {
```
holzeis, line 56: ❓ do we want to keep the sql queries here? I started adding them to the cfd file, but this is again growing quite big, so I wouldn't mind putting the queries somewhere else. However, adding it to the db file will bloat the db again.

Should we create a dedicated cfd module with the following structure?

```
open.rs
settles.rs
load_cfds.rs
insert_cfd.rs
```

holzeis: see bf6d650c65ed8fc08e2f0aca252b13f1e89cb090

lib/models/cfd.dart
```diff
@@ -0,0 +1,57 @@
...
+
+enum CfdStatus {
+  open,
+  closed,
+  failed,
+}
+
+extension CfdStatusExtension on CfdStatus {
+  static const displays = {
+    CfdStatus.open: "Opem",
```
luckysori: 🔧 Should be `Open`.

lib/models/order.dart
```diff
@@ -24,24 +24,19 @@ extension TradingPairExtension on TradingPair {
...
-    OrderStatus.open: "Open",
-    OrderStatus.closed: "Closed",
-    OrderStatus.failed: "Failed",
   };
 
   String get display => displays[this]!;
 }
 
 class Order {
+  // TODO: This id is likely not needed, because the `Order` will not be persisted (within the sc
```
luckysori: 🙃 Remove either the ID or the TODO?

lib/cfd_trading/cfd_offer.dart
```diff
@@ -43,9 +41,8 @@ class _CfdOfferState extends State<CfdOffer> {
         expiry: DateTime.now(),
         liquidationPrice: 13104,
         openPrice: 19100,
-        pl: Amount(Random().nextInt(20000) + -10000),
         quantity: 100,
-        estimatedFees: Amount(-4));
+        estimatedOpeningFees: Amount(-4));
```
luckysori: ❓ Could we just remove the opening fee for the tournament?

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -114,18 +110,16 @@ class _CfdOrderDetailState extends State<CfdOrderDetail> {
                               visible: confirm,
                               child: ElevatedButton(
                                   onPressed: () async {
-                                    // mock cfd has been closed.
-                                    Timer(const Duration(seconds: 2), () {
-                                      order.status = OrderStatus.closed;
-                                      cfdTradingService.persist(order);
-                                    });
+                                    // TODO: I think we don't need to update the UI state here, because it will be updated through polling the Cfd data from the backend
```
bonomat, line 114: Can we still do it, or at least trigger it from here? Otherwise the UX is not ideal because the user does an action and it takes a while before he sees it being changed in the frontend.

lib/models/cfd.dart
```diff
@@ -0,0 +1,57 @@
...
+  String get display => displays[this]!;
+}
+
+class Cfd {
+  String customOutputDetailId;
+
+  CfdStatus status;
+
+  int liquidationPrice;
+  int openPrice;
```
bonomat: Not sure if we should bother but with `int` you will loose the decimal precision.

lib/models/order.dart
bonomat: Todo is incomplete.

luckysori (approved):
LGTM, although I didn't do a very thorough review given the time constraints.

rust/migrations/202211170900_create_cfd_table.sql
```diff
@@ -5,15 +5,15 @@ CREATE TABLE IF NOT EXISTS cfd_state (
 INSERT INTO
     cfd_state (id, state)
 VALUES
```
bonomat, line 7: If you change the sql file you likely will need to run the migration again or they get out of sync.

lib/cfd_trading/cfd_offer.dart
```diff
@@ -56,8 +56,6 @@ class _CfdOfferState extends State<CfdOffer> {
     final margin = Amount(250000).display(currency: Currency.btc).value;
     // TODO: calculate expiry
     final expiry = DateFormat('dd.MM.yy-kk:mm').format(DateTime.now());
-    // TODO: calculate funding rate
-    final fundingRate = Amount(200).display(currency: Currency.btc).value;
```
bonomat: That's interesting. Imho we still need to charge some form of a _fee_ for the time the position is open. No need to action this for now. Just want to bring it up.

bonomat (approved):
Looks good to me.

rust/src/wallet.rs
```diff
@@ -254,7 +254,7 @@ impl Wallet {
     }
 }
 
-fn get_wallet() -> Result<MutexGuard<'static, Wallet>> {
+pub fn get_wallet() -> Result<MutexGuard<'static, Wallet>> {
```
klochowicz, line 257: nit: it will be safer if we won't leak the `MutexGuard` outside of this file - I will remove the `pub` in a follow-up. :)

klochowicz (approved):
thanks for all the work!

lib/cfd_trading/cfd_offer.dart
luckysori: Oh yeah, we still want to charge a fee. But for this early version I think it just makes things a bit more complicated than necessary.

lib/models/cfd.dart
da-kami: This was changed to `double`, we should have probably squashed a few commits here 😬

lib/cfd_trading/cfd_order_detail.dart
da-kami: See https://github.com/itchysats/10101/pull/231/commits/9d86a591083fc3bc1e349d0a7442fb4d09ac8277

lib/cfd_trading/cfd_offer.dart
da-kami: It was removed as far as I could see, sry should have squashed some commits :)

PR #233, opened 2022-11-17
feat: publish maker docker on each push to master

bonomat:
Resolves #205 

Note: the binary in the container is throwing an error at the moment.

```
Error: Cannot assign requested address (os error 99)
```

This might be a problem of how we listen for rocket. I'll create a ticket for that.


holzeis (approved):
👍

Issue #217, opened 2022-11-16
Parse command line arguments to configure maker


holzeis:
similar to how we have done it in itchysats using https://docs.rs/clap/latest/clap/

PR #237, opened 2022-11-17
chore: make maker listen on `0.0.0.0:8000`

### LINKED ISSUE #234: Make rocket listen on 0.0.0.0:8000
Maker should listen on `0.0.0.0:8000`

<details>
<summary>Outdated</summary>

> 
> https://github.com/itchysats/10101/pull/233 builds a docker container but the binary fails at the moment. This could be because of rocket. 
> 
> ```
> Error: Cannot assign requested address (os error 99)
> ```
> 
> To solve this I would make maker listen on 0.0.0.0:8000 instead of `localhost`.

</details>

bonomat:
resolves #234

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -39,7 +39,11 @@ async fn main() -> Result<()> {
 
     let (_, quote_receiver) = bitmex::subscribe()?;
 
-    let mission_success = rocket::build()
+    let figment = rocket::Config::figment()
+        .merge(("address", "0.0.0.0"))
+        .merge(("port", 8000));
+
+    let mission_success = rocket::custom(figment)
         .mount(
             "/api",
             rocket::routes![routes::get_offer, routes::post_force_close_channel],


klochowicz (approved):

maker/src/main.rs
```diff
@@ -39,7 +39,11 @@ async fn main() -> Result<()> {
 
     let (_, quote_receiver) = bitmex::subscribe()?;
 
-    let mission_success = rocket::build()
+    let figment = rocket::Config::figment()
+        .merge(("address", "0.0.0.0"))
```
klochowicz, line 43: perhaps a comment stating that we need this for Docker would be useful - might be useful for next projects :)

klochowicz (approved):
LGTM!

maker/src/main.rs
bonomat: I 'll overwrite this when doing  #217 and let the user pass it in.

Issue #228, opened 2022-11-17
Add continuous deployment to the maker


holzeis:
Automatically deploy the maker (upon merge to master) to the testnet vm.

bonomat:
For simplicity reasons I would use the [watchtower](https://github.com/containrrr/watchtower) again which simply polls regularly to see if a new container is available.

PR #239, opened 2022-11-17
Add maker cli

### LINKED ISSUE #217: Parse command line arguments to configure maker

bonomat:
Once this is in we can do #228 

Resolves #217

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,6 +423,43 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap"
+version = "4.0.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


.dockerignore
```diff
@@ -1,2 +1,3 @@
 *
 !target/release/maker
+!target/x86_64-unknown-linux-musl/release/maker
```
bonomat, line 3: was not on purpose

klochowicz: what Docker container we will run it in? with `musl`, everything libc should be statically linked, so we could use a minimal one I guess ;)

klochowicz (approved):
really cool!

.dockerignore
bonomat: I just needed this to be able to cross-compile locally :) We are using the standard linux target in our container

PR #241, opened 2022-11-18
fix: Add a timeout to reqwest client

klochowicz:
Sending HTTP requests and awaiting response without a timeout is needlessly risky.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -50,6 +50,8 @@ static MAKER_PORT_HTTP: u64 = 8000;
 // Maker PK is derived from our checked in regtest maker seed
 static MAKER_PK: &str = "02cb6517193c466de0688b8b0386dbfb39d96c3844525c1315d44bd8e108c08bc1";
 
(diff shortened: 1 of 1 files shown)


rust/src/wallet.rs
```diff
@@ -362,7 +364,7 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()>
 
     let address_to_fund = get_address()?;
 
-    let client = reqwest::Client::new();
+    let client = reqwest::Client::builder().timeout(TCP_TIMEOUT).build()?;
```
holzeis, line 367: We should probably also do that for get offer.

klochowicz: happy to accept a PR for that :)

holzeis: I was starting with it, but I was wondering how we want to continue with our constants? For now I can use it from the wallet.rs, but I don't think it should stay there.

holzeis: https://github.com/itchysats/10101/pull/249

PR #242, opened 2022-11-18
fix: use correct target binary

bonomat:
This sneaked in 🙈


bonomat:
Merging this rightaway, no need to block CI.

PR #243, opened 2022-11-18
Print taker app node ID on start-up

luckysori:


klochowicz (approved):
thanks, that's a helpful addition :)

PR #246, opened 2022-11-18
fix: display correct sent amount

bonomat:
The wallet returns us a transaction detail which includes the sent amount and received amount. To calculate how much we really sent someone else we need to do some math as the `sent` amount includes the whole amount including the one we receive back in the return.
1. subtract the received amoumt
2. subtract the transaction fee

The remaining is the exact amount we sent someone.


da-kami (approved):
Thanks for fixing this!

luckysori (approved):
Looked at the formula and it LGTM!

PR #251, opened 2022-11-18
feat: Support paying lightning invoices

klochowicz:
In order to test, I have added some basic API to generate an invoice on the maker side.

TODO(follow-up): Improve the invoice generation API and hook it up into 10101.

### DIFF
--- a/lib/wallet/send.dart
+++ b/lib/wallet/send.dart
@@ -1,17 +1,85 @@
+import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
+import 'package:go_router/go_router.dart';
+import 'package:ten_ten_one/ffi.io.dart';
 
-class Send extends StatelessWidget {
+class Send extends StatefulWidget {
   const Send({Key? key}) : super(key: key);
 
   static const route = '/' + subRouteName;
   static const subRouteName = 'send';
 
   @override
+  State<Send> createState() => _SendState();
+}
+
+class _SendState extends State<Send> {
+  String invoice = "";
+
+  @override
+  void initState() {
+    super.initState();
+  }
+
+  @override
   Widget build(BuildContext context) {
     return Scaffold(
         appBar: AppBar(
-          title: const Text('Send Payment'),
+          title: const Text('Send Lightning Invoice'),
         ),
-        body: ListView(children: const [Center(child: Text("Send Payment"))]));
+        body: SafeArea(
+            child: Padding(
+                padding: const EdgeInsets.all(20.0),
+                child: Column(children: [
+                  Column(crossAxisAlignment: CrossAxisAlignment.start, children: [
+                    const Text("Invoice", style: TextStyle(color: Colors.grey)),
+                    TextFormField(
+                      initialValue: "",
+                      keyboardType: TextInputType.text,
+                      onChanged: (text) {
+                        setState(() {
+                          invoice = text;
+                        });
+                      },
+                    ),
+                    const SizedBox(
+                      height: 10.0,
+                    ),
+                  ]),
+                  Expanded(
+                    child: Container(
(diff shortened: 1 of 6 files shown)


klochowicz:
I've tested it by creating an invoice from the maker, and pasted it in into the taker.
result after sending a couple invoices: 
image

note: I have managed to send the same invoice twice, which is probably not ideal.

lib/wallet/send.dart
```diff
@@ -1,17 +1,85 @@
...
+  void initState() {
+    super.initState();
+  }
+
   @override
   Widget build(BuildContext context) {
     return Scaffold(
         appBar: AppBar(
-          title: const Text('Send Payment'),
+          title: const Text('Send Lightning Invoice'),
```
bonomat, line 29: 🙃
That reads weird. Imho you are `paying` an invoice, not `sending` it.

lib/wallet/send.dart
```diff
@@ -1,17 +1,85 @@
...
+                      margin: const EdgeInsets.fromLTRB(0, 0, 20, 30),
+                      child: Column(
+                        mainAxisAlignment: MainAxisAlignment.end,
+                        children: [
+                          Row(
+                            mainAxisAlignment: MainAxisAlignment.end,
+                            children: [
+                              ElevatedButton(
+                                  onPressed: () async {
+                                    FLog.info(text: "Sending invoice" + invoice);
```
bonomat, line 61: Same as above. `Paying` and invoice

maker/src/routes.rs
```diff
@@ -119,3 +121,22 @@ pub async fn post_open_channel(
 
     Ok(Json(OpenChannelResponse { funding_txid }))
 }
+
+#[rocket::post("/invoice/send/<invoice>")]
```
bonomat, line 125: I'd also change this `send` to `pay`.
Note, this API is open to the public. I don't mind for the tournament but thought to point it out.

bonomat (approved):
Looks good. I'd just change the naming a bit.

da-kami:
> I've tested it by creating an invoice from the maker, and pasted it in into the taker.
result after sending a couple invoices:

One thing I noticed is odd in the screenshot is, that it says `Payment Sent` but then it displays `+10` - I think it should be `-10` if we sent it?

bonomat:
> > I've tested it by creating an invoice from the maker, and pasted it in into the taker.
> > result after sending a couple invoices:
> 
> One thing I noticed is odd in the screenshot is, that it says `Payment Sent` but then it displays `+10` - I think it should be `-10` if we sent it?

You might need to adjust it in the frontend. Same for the BTC payments.

rust/src/wallet.rs
```diff
@@ -551,6 +579,7 @@ pub enum Flow {
     Outbound,
 }
 
+// TODO: Remove this? Seems to be exactly the same as HTLCStatus
```
da-kami, line 582: I would not bother :)

da-kami (approved):

rust/src/wallet.rs
klochowicz: let's keep this until the tournament ends, but then probably needs to go, as this is just adding confusion for no reason :p

klochowicz:
@bonomat I've created a github issue to make sure we can align on the naming after the tournament: https://github.com/itchysats/10101/issues/258

maker/src/routes.rs
klochowicz: I've created a new ticket out of this so we won't forget after the tournament.

klochowicz: > I'd also change this `send` to `pay`. Note, this API is open to the public. I don't mind for the tournament but thought to point it out.

that's true. I don't think adding authentication when only using testnet makes much sense, so I didn't add it yet.
However, I added a ticket to ensure we will not forget about it later: 
https://github.com/itchysats/10101/issues/259

PR #255, opened 2022-11-18
Add calculations

holzeis:
Replaces mocked data with actual calculated data.

- [x] margin calcuation
- [x] expiry calcuation
- [x] pnl calcuation
- [x] estimated fees

Please have a close eye on the calculations. E.g. I did not include any fees in the online calculation. Not sure if everything is correct. It looked good to me, though.

Side note, we should start writing tests again after the tournament. 😅

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -51,7 +51,7 @@ class _CfdOfferState extends State<CfdOffer> {
 
     final liquidationPrice = formatter.format(order.calculateLiquidationPrice());
     final expiry = DateFormat('dd.MM.yy-kk:mm')
-        .format(DateTime.fromMillisecondsSinceEpoch(order.calculateExpiry()));
+        .format(DateTime.fromMillisecondsSinceEpoch((order.calculateExpiry() * 1000)));
     final margin = formatter.format(order.calculateMargin());
 
     return Scaffold(
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -40,7 +40,7 @@ class _CfdOfferState extends State<CfdOffer> {
     final fmtIndex = formatter.format(offer.index);
 
     cfdTradingService.draftOrder ??= Order(
-        openPrice: offer.index,
+        openPrice: offer.ask,
         quantity: 100,
         leverage: 2,
         contractSymbol: ContractSymbol.BtcUsd,
@@ -52,7 +52,7 @@ class _CfdOfferState extends State<CfdOffer> {
     final liquidationPrice = formatter.format(order.calculateLiquidationPrice());
     final expiry = DateFormat('dd.MM.yy-kk:mm')
         .format(DateTime.fromMillisecondsSinceEpoch((order.calculateExpiry() * 1000)));
-    final margin = formatter.format(order.calculateMargin());
+    final margin = formatter.format(order.marginTaker());
 
     return Scaffold(
       body: ListView(padding: const EdgeInsets.only(left: 25, right: 25), children: [
@@ -71,6 +71,7 @@ class _CfdOfferState extends State<CfdOffer> {
             onChange: (position) {
               setState(() {
                 order.position = position!;
+                order.openPrice = Position.Long == position ? offer.ask : offer.bid;
               });
             },
             value: order.position),
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -46,12 +46,12 @@ class _CfdOfferState extends State<CfdOffer> {
       leverage: 2,
       position: Position.Long,
       contractSymbol: ContractSymbol.BtcUsd,
+      bridge: api,
     );
 
     order = cfdTradingService.draftOrder!;
 
-    final liquidationPrice = api.calculateLiquidationPrice(
-        initialPrice: order.openPrice, leverage: order.leverage, position: order.position);
+    final liquidationPrice = order.calculateLiquidationPrice();
 
     // TODO: calcualte margin
     final margin = Amount(250000).display(currency: Currency.sat).value;
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -51,10 +51,7 @@ class _CfdOfferState extends State<CfdOffer> {
     order = cfdTradingService.draftOrder!;
 
     final liquidationPrice = api.calculateLiquidationPrice(
-        initialPrice: order.openPrice,
-        leverage: order.leverage,
-        contractSymbol: order.contractSymbol,
-        position: order.position);
+        initialPrice: order.openPrice, leverage: order.leverage, position: order.position);
 
     // TODO: calcualte margin
     final margin = Amount(250000).display(currency: Currency.sat).value;
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -9,7 +9,6 @@ import 'package:ten_ten_one/cfd_trading/cfd_order_confirmation.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
 import 'package:ten_ten_one/cfd_trading/position_selection.dart';
-import 'package:ten_ten_one/models/amount.model.dart';
 import 'package:ten_ten_one/utilities/divider.dart';
 import 'package:ten_ten_one/utilities/dropdown.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
@@ -41,24 +40,19 @@ class _CfdOfferState extends State<CfdOffer> {
     final fmtIndex = formatter.format(offer.index);
 
     cfdTradingService.draftOrder ??= Order(
-      openPrice: offer.index,
-      quantity: 100,
-      leverage: 2,
-      position: Position.Long,
-      contractSymbol: ContractSymbol.BtcUsd,
-      bridge: api,
-    );
+        openPrice: offer.index,
+        quantity: 100,
+        leverage: 2,
+        contractSymbol: ContractSymbol.BtcUsd,
+        position: Position.Long,
+        bridge: api);
 
     order = cfdTradingService.draftOrder!;
 
-    final liquidationPrice = order.calculateLiquidationPrice();
-
-    // TODO: calcualte margin
-    final margin = Amount(250000).display(currency: Currency.sat).value;
-
-    // TODO: We also have to pass this to rust, otherwise it won't align with the cfd details later
-    final now = DateTime.now();
-    final expiry = DateTime(now.year, now.month, now.day + 1);
+    final liquidationPrice = formatter.format(order.calculateLiquidationPrice());
+    final expiry = DateFormat('dd.MM.yy-kk:mm')
+        .format(DateTime.fromMillisecondsSinceEpoch(order.calculateExpiry()));
+    final margin = formatter.format(order.calculateMargin());
 
     return Scaffold(
       body: ListView(padding: const EdgeInsets.only(left: 25, right: 25), children: [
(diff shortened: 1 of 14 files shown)


rust/src/api.rs
```diff
@@ -192,37 +186,55 @@ pub fn get_seed_phrase() -> Result<Vec<String>> {
...
+// Note, this implementation has to be on the api level as otherwise it wouldn't be generated
+// through frb. TODO: Provide multiple rust targets to the code generation so that this code can be
+// nicer structured.
+impl Order {
+    pub fn calculate_margin(&self) -> SyncReturn<f64> {
+        let margin = match self.contract_symbol {
+            ContractSymbol::BtcUsd => {
+                self.quantity as f64 / (self.open_price * self.leverage as f64)
+            }
+            ContractSymbol::EthUsd => todo!(),
```
luckysori, line 215: 🤔 Why do we have an ETHUSD variant?!?

rust/src/api.rs
```diff
@@ -192,37 +186,55 @@ pub fn get_seed_phrase() -> Result<Vec<String>> {
...
+            }
+            (ContractSymbol::BtcUsd, Position::Short) => {
+                calc::inverse::calculate_short_liquidation_price(leverage, initial_price)
+            }
+            (ContractSymbol::EthUsd, Position::Long) => {
+                calc::quanto::bankruptcy_price_long(initial_price, leverage)
+            }
+            (ContractSymbol::EthUsd, Position::Short) => {
+                calc::quanto::bankruptcy_price_short(initial_price, leverage)
+            }
```
luckysori, line 247: 🤔 Surely not needed.

lib/cfd_trading/cfd_order_confirmation.dart
```diff
@@ -73,28 +73,30 @@ class CfdOrderConfirmation extends StatelessWidget {
...
                               onPressed: () async {
-                                // TODO: Plug in actual state changes
-
                                 try {
-                                  // TODO: Don't hardcode the taker amount
                                   await api.openCfd(order: order);
                                   FLog.info(text: 'OpenCfd returned successfully');
 
+                                  // refreshing cfd list after cfd has been opened
+                                  api.listCfds();
```
luckysori, line 98: ❓ This has a side-effect? We're not doing anything with the returned value.

luckysori: It does. The name is a bit confusing. Something like `syncCfds` or `updateCfds` might be clearer.

lib/cfd_trading/cfd_trading_change_notifier.dart
```diff
@@ -28,10 +27,14 @@ class CfdTradingChangeNotifier extends ChangeNotifier {
     super.notifyListeners();
   }
 
+  Future<void> listCfds() async {
+    cfds = await api.listCfds();
+    super.notifyListeners();
+  }
```
luckysori, line 33: It's a bit annoying that we have to call this every time we affect the list of CFDs. It would be nicer if it were just updated periodically (like we do below) and also every time we navigate to the page that lists the CFDs.

lib/cfd_trading/cfd_offer.dart
```diff
@@ -41,27 +40,19 @@ class _CfdOfferState extends State<CfdOffer> {
...
-      leverage: 2,
-      position: Position.Long,
-      contractSymbol: ContractSymbol.BtcUsd,
-    );
+        openPrice: offer.index,
+        quantity: 100,
+        leverage: 2,
+        position: Position.Long,
+        contractSymbol: ContractSymbol.BtcUsd,
+        bridge: api);
```
luckysori, line 48: ❓ What's this for?

luckysori: It's never used in this patch AFAICT.

rust/src/api.rs
```diff
@@ -245,6 +252,54 @@ impl Order {
...
+    }
+
+    pub fn calculate_payout_at_price(&self, closing_price: f64) -> Result<SyncReturn<f64>> {
+        match self.contract_symbol {
+            ContractSymbol::BtcUsd => {
+                let long_margin = self.calculate_margin().0;
+                let short_margin = long_margin;
+                // TODO: assuming the maker always goes with the same leverage as the taker
+                // which might not be the case.
+                let total_margin = long_margin + short_margin;
```
luckysori, line 277: 📣 Wrong assumption 😛 I think we just need to get the margins like this:

```rust
let taker_margin = self.calculate_margin().0;
let maker_margin = taker_margin * self.leverage;
```

luckysori (commented):
Not super thorough review, but it LGTM apart from a couple small things.

rust/src/api.rs
```diff
@@ -192,37 +186,55 @@ pub fn get_seed_phrase() -> Result<Vec<String>> {
...
-) -> SyncReturn<f64> {
-    let initial_price = Decimal::try_from(initial_price).expect("Price to fit");
+// Note, this implementation has to be on the api level as otherwise it wouldn't be generated
+// through frb. TODO: Provide multiple rust targets to the code generation so that this code can be
+// nicer structured.
+impl Order {
+    pub fn calculate_margin(&self) -> SyncReturn<f64> {
+        let margin = match self.contract_symbol {
+            ContractSymbol::BtcUsd => {
+                self.quantity as f64 / (self.open_price * self.leverage as f64)
```
bonomat: Checked the calc, looks good.

bonomat (approved):
I think this is good to go 👍

rust/src/api.rs
luckysori: I've removed ETHUSD-related stuff with approval from the rest of the team.

lib/cfd_trading/cfd_order_confirmation.dart
luckysori: I went for `refreshCfds` and also fixed the commit, because here we were only calling `api.listCfds()`, which returns the list which we were ignoring. Instead we had to call:

```dart
await cfdTradingService.refreshCfdList();
```

lib/cfd_trading/cfd_trading_change_notifier.dart
luckysori: Not dealing with it right now, because we want to move away from polling if possible.

lib/cfd_trading/cfd_offer.dart
luckysori: I think I understand it now.

rust/src/api.rs
luckysori: Amended in 16ec2338fe65b4b5578ceddcd3775642589c6f6b.

lib/cfd_trading/cfd_order_confirmation.dart
luckysori: Fixed in fd57d7499d73a3cac748ec27fb0ee825ae0d998c.

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -25,23 +27,43 @@ class CfdOrderDetail extends StatefulWidget {
...
+    final cfdOffersChangeNotifier = context.watch<CfdOfferChangeNotifier>();
+    final offer = cfdOffersChangeNotifier.offer ?? Offer(bid: 0, ask: 0, index: 0);
+
     Cfd cfd = widget.cfd!;
 
     final openPrice = formatter.format(cfd.openPrice);
     final liquidationPrice = formatter.format(cfd.liquidationPrice);
-    final estimatedFees = Amount(-4).display(currency: Currency.sat, sign: true).value;
-    final margin = Amount(200).display(currency: Currency.sat).value;
+    final margin = Amount.fromBtc(cfd.margin).display(currency: Currency.btc).value;
```
da-kami, line 58: I think @bonomat tried to display everything as `sat` so we don't mix sat and btc values. 
We might want to change this to `sat` as well.

da-kami, line 58:
```suggestion
    final margin = Amount.fromBtc(cfd.margin).display(currency: Currency.sat).value;
```

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -25,23 +27,43 @@ class CfdOrderDetail extends StatefulWidget {
...
+    final estimatedFees = Amount(fastestFee).display(currency: Currency.btc).value;
```
da-kami, line 59:
```suggestion
    final estimatedFees = Amount(fastestFee).display(currency: Currency.sat).value;
```

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -116,12 +138,21 @@ class _CfdOrderDetailState extends State<CfdOrderDetail> {
                                   onPressed: () async {
                                     try {
                                       await api.settleCfd(takerAmount: 40000, makerAmount: 20000);
+                                      FLog.info(text: "Successfully settled cfd.");
+
+                                      // refreshing cfd list after cfd has been closed
+                                      await cfdTradingService.refreshCfdList();
+
+                                      context.go(CfdTrading.route);
```
da-kami, line 146: 🛠️ We trigger the router in the code below on line 166 - I don't think we should jump here, because that would mean that we introduce a "flicker" because we route here, then change index and route again 🙈

lib/cfd_trading/cfd_overview.dart
```diff
@@ -22,8 +23,10 @@ class CfdOverview extends StatefulWidget {
 class _CfdOverviewState extends State<CfdOverview> {
   @override
   Widget build(BuildContext context) {
+    final cfdOffersChangeNotifier = context.watch<CfdOfferChangeNotifier>();
```
da-kami, line 26: 🛠️ This is a duplicate of line `27`! We should just remove the one named `service`

rust/src/api.rs
```diff
@@ -196,37 +200,98 @@ pub fn send_lightning_payment(invoice: String) -> Result<()> {
...
+            let quantity = Decimal::from(self.quantity);
+
+            let uncapped_pnl = (quantity / opening_price) - (quantity / closing_price);
+            let uncapped_pnl = uncapped_pnl
+                .round_dp_with_strategy(8, rust_decimal::RoundingStrategy::MidpointAwayFromZero);
+            uncapped_pnl
+                .to_f64()
+                .context("Could not convert Decimal to f64")?
+        };
+        // TODO: check if we need to consider tx fees here.
```
da-kami, line 289: For now we could say that tx-fees are just not part of this calc...

rust/src/offer.rs
```diff
@@ -23,3 +25,31 @@ pub async fn get_offer() -> Result<Offer> {
...
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeRecommendation {
+    pub fastest_fee: u32,
+    pub half_hour_fee: u32,
+    pub hour_fee: u32,
+    pub minimum_fee: u32,
+}
+
+pub async fn get_fee_recommendation(network: Network) -> Result<FeeRecommendation> {
```
da-kami, line 38: 🛠️ We have the possibility to use Electrum for this! I added this before:

```
let estimated_fee_rate = self
            .lightning
            .wallet
            .estimate_fee(ConfirmationTarget::Normal)
            .map_err(|_| anyhow!("Failed to estimate fee"))?;
```

I think a dependency on mempool is not needed here, and just makes everything even more brittle! we cannot easily know if we hit quotas with their API.

da-kami (commented):
I added some 🛠️es 😬
I'll have another look once we actioned these :)

lib/cfd_trading/cfd_order_detail.dart
luckysori: Seems legit.

lib/cfd_trading/cfd_overview.dart
luckysori: Don't think so. One is for CFDs and the other is for the offer.

da-kami: Fair enough, I mixed that up!

rust/src/offer.rs
luckysori: Done!

PR #261, opened 2022-11-21
Ignore maker funding transaction in bitcoin tx history

### LINKED ISSUE #232: Hide maker's funding transaction in tx-history of taker
This is a follow-up for https://github.com/itchysats/10101/issues/199

This part is to be added:

I think the taker's user interface should not  show this transaction otherwise that might be confusing, we might want to filter this address when displaying the history.

Simplest way is to store the `txid` in a database table of `txid`s to be ignored and before sending the tx history out filter for those txids to be ignored removing the entries :)

Note: There are some pitfalls where an address could potentially be used for both maker funds and taker funds if the taker manages to deposit funds to the same address. I don't think this is possible because we always use the next unused address, but not sure our process is sound enough against that. Would ignore that problem :)

For the balance we could just ignore subtracting the value and accept that it might change for a few sec.
The maker should send out the funds right away and we react to the tx in mempool and chain the channel open on top, so this should not really matter for the balance.

da-kami:
resolves #232

We save the txid of the funding tx and filter it when creating the history.

### DIFF
--- a/rust/migrations/20221121024641_ignore_maker_funding_tx_table.sql
+++ b/rust/migrations/20221121024641_ignore_maker_funding_tx_table.sql
@@ -1,0 +1,5 @@
+-- Store the maker's funding transaction id so we can ignore it when creating the bitcoin tx history
+CREATE TABLE IF NOT EXISTS ignore_txid (
+    id INTEGER PRIMARY KEY,
+    txid TEXT NOT NULL
+);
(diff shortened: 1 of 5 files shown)


bonomat (approved):

bonomat:
LGTM :)

luckysori (approved):
Nice!

da-kami:
I added another commit so we don't see the maker's amount - looks like this now:

image

klochowicz (approved):

luckysori (approved):

PR #262, opened 2022-11-21
Support more than one channel

klochowicz:
Remove some arbitrary constraints, such as:
- add up the balance in all open channels
- when opening CFD, look at first open channel with *maker*

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -13,8 +13,8 @@ use rust_decimal_macros::dec;
 use ten_ten_one::wallet::force_close_channel;
 use ten_ten_one::wallet::get_address;
 use ten_ten_one::wallet::get_balance;
+use ten_ten_one::wallet::get_channel_manager;
 use ten_ten_one::wallet::get_invoice;
-use ten_ten_one::wallet::get_lightning_channels;
 use ten_ten_one::wallet::get_node_id;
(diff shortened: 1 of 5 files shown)


klochowicz:
note: I was able to successfully open a channel with two takers, but for some reason, still have troubles with opening CFD.

maker/src/routes.rs
```diff
@@ -140,3 +141,17 @@ pub async fn get_new_invoice() -> Result<(), HttpApiProblem> {
...
+    let list = get_channel_manager()
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Failed to create lightning invoice")
+                .detail(format!("{e:#}"))
+        })?
+        .list_channels();
+
+    tracing::info!(?list, "Open channels: {}", list.len());
+    Ok(())
```
luckysori, line 156: ❓ Why not returning the list instead of logging?

luckysori (approved):

maker/src/routes.rs
klochowicz: good question! 
I chose this because `ChannelDetails` doesn't support `Serialize` out of the box, and I figured out that implementing a custom `Writer` or our own type would be an overkill right now, esp. when we're not really using it outside of testing.

PR #264, opened 2022-11-21
Rename app to 10101 everywhere

klochowicz:


luckysori (approved):

linux/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.11)
 project(runner LANGUAGES CXX)
 
-set(BINARY_NAME "with_flutter")
-set(APPLICATION_ID "com.example.with_flutter")
+set(BINARY_NAME "10101")
+set(APPLICATION_ID "com.example.10101")
```
holzeis, line 5: not sure if this will cause any issues (since not a valid package name), but go ahead with the merge if its working for you.

holzeis (approved):

linux/CMakeLists.txt
klochowicz: in this PR the package name stayed as `ten_ten_one`. I have changed it everywhere else.
Haven't tested windows / mobile platforms, but on macos works just fine.

PR #274, opened 2022-11-21
Calculate settlement amounts

### LINKED ISSUE #219: Do not hard-code settlement amounts
With #207 we introduced the ability to actually settle CFDs (removing the custom output from the Lightning channel). But the values used are hard-coded in the frontend.

### Rough design
1. User navigates to CFD order details page (data loaded from storage, see #201).
2. User presses the `settle` button.
3. We use some of the persisted values together with the current price to call a `calculate_payouts_function` which acts as the payout curve.
4. The function returns `maker_amount` and `taker_amount` which can replace the hard-coded values.

https://github.com/itchysats/10101/blob/a91a50c3df483a95fa61e265d4671bb64286622f/lib/cfd_trading/cfd_order_detail.dart#L124

holzeis:
This PR calculates the settlement amounts and updates the cfd state in the database. Still not a 100% sure if all the calculations are correct, but at least the maker + taker amount add up to the total margin - so that the settlement goes through 🥳 

resolves #219

### DIFF
--- a/rust/src/cfd.rs
+++ b/rust/src/cfd.rs
@@ -56,6 +56,7 @@ pub struct Cfd {
     pub quantity: i64,
     pub expiry: i64,
     pub open_price: f64,
+    pub close_price: Option<f64>,
     pub liquidation_price: f64,
     pub margin: f64,
 }
@@ -162,7 +163,7 @@ pub async fn open(order: &Order) -> Result<()> {
     Ok(())
 }
 
-pub async fn settle(cfd: Cfd, offer: Offer) -> Result<()> {
+pub async fn settle(cfd: &Cfd, offer: &Offer) -> Result<()> {
     // TODO: need to derive an order from the cfd as dependent functions are only available on the
     // order eventually the order should probably be included in the cfd.
     let order = cfd.derive_order();
@@ -191,7 +192,7 @@ pub async fn settle(cfd: Cfd, offer: Offer) -> Result<()> {
 
     let channel_manager = wallet::get_channel_manager()?;
 
-    let custom_output_id = base64::decode(cfd.custom_output_id)?;
+    let custom_output_id = base64::decode(&cfd.custom_output_id)?;
     let custom_output_id: [u8; 32] = custom_output_id
         .try_into()
         .expect("custom output ID to be 32 bytes long");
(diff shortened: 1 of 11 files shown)


rust/src/cfd.rs
```diff
@@ -145,8 +149,37 @@ pub async fn open(order: &Order) -> Result<()> {
...
+    let taker_amount_sats = taker_amount_sats
+        .round_dp_with_strategy(8, rust_decimal::RoundingStrategy::MidpointAwayFromZero);
+
+    let maker_amount_sats = total_margin_sats - taker_amount_sats;
+
+    assert_eq!(total_margin_sats, maker_amount_sats + taker_amount_sats);
+
+    tracing::info!(
+        "Settling CFD with taker amount {taker_amount_sats} and maker amount {maker_amount_sats}"
+    );
```
bonomat, line 197: A nice feature of tracing is to add the variables as field names. This makes log analyzis easier as the log message always remains the same and the variables are extra, e.g.

```suggestion
    tracing::info!(taker_amount_sats, maker_amount_sats, 
        "Settling CFD"
    );
```

 in json format this will look something like this: 

```json
{
timestamp: xxx
message: "Settling CFD",
fields: {
  taker_amount_sats: yyy,
  maker_amount_sats: zzz,
}
}
```

rust/src/cfd.rs
```diff
@@ -145,8 +149,37 @@ pub async fn open(order: &Order) -> Result<()> {
...
+        Position::Long => offer.bid,
+        Position::Short => offer.ask,
+    };
+
+    let taker_amount_sats =
+        Decimal::try_from(order.calculate_profit(price)?.0)? * Decimal::from(100_000_000);
+    // add profit or substract loss from takers margin. we could calculate the payout here as well,
+    // but this is already implicitly included in the profit.
+    let taker_amount_sats = taker_margin_sats + taker_amount_sats;
+    // TODO: adding this only as a safe guard as this should not happen. I guess it could happen
```
bonomat, line 184: No need to add a TODO here, I think it's OK to have this comment and ensure that the amount does not go negative.

rust/src/cfd.rs
```diff
@@ -145,8 +149,37 @@ pub async fn open(order: &Order) -> Result<()> {
...
+    // though until we automatically close on liquidation price.
+    let taker_amount_sats = taker_amount_sats.max(Decimal::ZERO);
+
+    let taker_amount_sats = taker_amount_sats
+        .round_dp_with_strategy(8, rust_decimal::RoundingStrategy::MidpointAwayFromZero);
```
bonomat, line 189: 🔧
If `taker_amount_sats` is indeed in sats we should round to `0`.

rust/src/cfd.rs
```diff
@@ -67,6 +70,7 @@ pub async fn open(order: &Order) -> Result<()> {
     let margin_taker_as_btc = order.margin_taker().0;
     // Convert to msats
     let margin_taker = (margin_taker_as_btc * 100_000_000.0 * 1000.0) as u64;
+    // TODO: is this correct? why do we multiply the leverage for the maker margin?
```
bonomat, line 87: The margin is depending on the contract value in BTC, i.e. `contract_value=quantity/price`.
To ensure the maker puts _enough_ margin for the taker to make a profit, we said his margin needs to be equal the `contract_value`, which ends up saying be the `leverage=1`, because, `margin = quantity / (open_price * leverage);` 

For the taker, he can chose a leverage for his margin, meaning, he only has to put up a fraction of the `contract_value`. If we turn the formulas around a bit we get: 
`contract_value=quantity/price` ->
`margin_taker=contract_value/leverage` -->
`contract_value=margin_taker*leverage` --> `margin_maker`.

rust/src/cfd.rs
```diff
@@ -183,10 +197,11 @@ pub async fn settle(order: Order, offer: Offer) -> Result<()> {
 
     let channel_manager = wallet::get_channel_manager()?;
 
-    let custom_output_id = *channel_manager
-        .custom_outputs()
-        .first()
-        .context("No custom outputs in channel")?;
+    let custom_output_id = base64::decode(cfd.custom_output_id)?;
```
bonomat: 🛑 wait, why does this work? 

The `custom_output_id` is randomly generated from the library.

bonomat (dismissed):
I think there is a bug in this PR.

rust/src/cfd.rs
luckysori: I think we are just reloading it from the database, where it is serialised as base64.

bonomat: oh, right, I misread it then :)

PR #289, opened 2022-11-22
feat: Generate lightning invoices

klochowicz:
- align the API to allow passing custom description when creating invoice
- Add a Flutter widget that allows creating an invoice. You can specify amount,
  expiry time, and add a custom description.
  It generates a QR code and allows you to copy invoice to clipboard.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -157,6 +157,8 @@ class _TenTenOneState extends State<TenTenOneApp> {
       await setupRustLogging();
 
       final appSupportDir = await getApplicationSupportDirectory();
+      FLog.info(text: "App data will be stored in: " + appSupportDir.toString());
+
       await api.initWallet(network: Network.Regtest, path: appSupportDir.path);
       await api.initDb(network: Network.Regtest, appDir: appSupportDir.path);
       await api.testDbConnection(); // TODO: Remove this call after testing DB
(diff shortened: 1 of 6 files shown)


klochowicz:
Uploading create_invoice.mp4…

bonomat (approved):
Nice, can the maker actually pay it?

klochowicz:
> Nice, can the maker actually pay it?

nope 😢 we need an outbound route from the maker to do so...
```
  {
    "status": 500,
    "title": "Failed to pay lightning invoice",
```

bonomat:
> > Nice, can the maker actually pay it?
> 
> nope 😢 we need an outbound route from the maker to do so...
> 
> ```
>   {
>     "status": 500,
>     "title": "Failed to pay lightning invoice",
> ```

This exists: the channel will have funding on both sides...

klochowicz:
> This exists: the channel will have funding on both sides...

I've updated the error log (somehow pasted without the last part, which was crucial) - seems there needs to be outbound route from the maker?

bonomat:
> > This exists: the channel will have funding on both sides...
> 
> I've updated the error log (somehow pasted without the last part, which was crucial) - seems there needs to be outbound route from the maker?

If you have opened a channel from the taker to the maker using our protocol then there is funding on the side of the maker and paying should be possible. I think the problem is somewhere else. 

I tried your branch locally and it looks like it is working but the payment remains in `pending` state. Might just be a frontend issue?

image

PR #290, opened 2022-11-22
chore: format payment history

### LINKED ISSUE #286: Improve Dashboard after receiving (pending transaction)
# Dashboard after receiving (pending transaction)

simulator_screenshot_577F35E6-BD4B-499F-85FE-0C30AF92A5D1

- `Bitcoin Received` should be online. Maybe change the text? 
- Every information should be on the same _level_, i.e. the tx history is on the lowest level and the `CFD Trading` on the highest which even has a shadow. 
- Add a date, when this transaction was confirmed

Potentially get inspired by Zeus: 
image

_Originally posted by @bonomat in https://github.com/itchysats/10101/discussions/273#discussioncomment-4193815_

    add time indicator (second row)
    Ellipse (overflow) on Text in the middle
    Reduce padding (in list as well as outer)

Service tiles:

    Remove shade, outline with theme color (? - see how that looks)

bonomat:
resolves #286

Before: 
Simulator Screen Shot - iPhone 14 - 2022-11-21 at 13 24 53

After: 
simulator_screenshot_2D7ACF80-FAE9-4D66-9EA1-6407CDDA16DC

### DIFF
--- a/lib/models/payment.model.dart
+++ b/lib/models/payment.model.dart
@@ -32,12 +32,12 @@ enum PaymentType {
 
 extension PaymentTypeExtension on PaymentType {
   static const displays = {
-    PaymentType.receiveOnChain: "Bitcoin Received",
-    PaymentType.sendOnChain: "Bitcoin Sent",
+    PaymentType.receiveOnChain: "You received",
+    PaymentType.sendOnChain: "You sent",
     PaymentType.channelOpen: "Channel Opened",
     PaymentType.channelClose: "Channel Closed",
-    PaymentType.send: "Payment Sent",
-    PaymentType.receive: "Payment Received",
+    PaymentType.send: "You sent",
+    PaymentType.receive: "You received",
     PaymentType.cfdOpen: "CFD Opened",
     PaymentType.cfdClose: "CFD Closed",
     PaymentType.sportsbetOpen: "Sports Bet Started",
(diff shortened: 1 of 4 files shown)


lib/models/payment.model.dart
```diff
@@ -32,12 +32,12 @@ enum PaymentType {
 
 extension PaymentTypeExtension on PaymentType {
   static const displays = {
-    PaymentType.receiveOnChain: "Bitcoin Received",
-    PaymentType.sendOnChain: "Bitcoin Sent",
+    PaymentType.receiveOnChain: "You received",
```
klochowicz, line 35: I'm not sure whether "You" is needed (here and elsewhere). consider just "Received"

lib/wallet/payment_history_list_item.dart
```diff
@@ -58,19 +64,42 @@ class PaymentHistoryListItem extends StatelessWidget {
...
-        crossAxisAlignment: CrossAxisAlignment.baseline,
-        textBaseline: TextBaseline.alphabetic,
-        children: [
-          AmountItem(text: amountDisplay.value, unit: AmountUnit.satoshi, iconColor: Colors.grey)
-        ],
-      ),
-    );
+    final date = data.status == PaymentStatus.pending
+        ? "pending"
+        : DateFormat.yMMMd()
```
klochowicz, line 69: can we add a bit more precision (hours and minutes?). I would like to see that in my payment history.

klochowicz (approved):
nice improvement!

for the future: something we haven't addressed at all are descriptions of lightning payments. We could display that in the second row as well.

lib/models/payment.model.dart
bonomat: I `you received` because it speaks to me :)

lib/wallet/payment_history_list_item.dart
bonomat: Done: 
image

PR #292, opened 2022-11-22
chore: Fix product strings

klochowicz:
Use `10101` as the product name consistently, including copyright notices.
I've also set `itchysats` as the company.

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -374,7 +374,7 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/Runner",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.example.frbExample;
+				PRODUCT_BUNDLE_IDENTIFIER = com.itchysats.10101;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				STRIP_STYLE = debugging;
 				SWIFT_OBJC_BRIDGING_HEADER = "Runner/Runner-Bridging-Header.h";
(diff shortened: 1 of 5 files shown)


android/app/build.gradle
```diff
@@ -51,7 +51,7 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "com.example.frb_example"
+        applicationId "com.itchysats.10101"
```
holzeis, line 54: Reading the guideline, referenced in the TODO above it seems this application id does not adhere to the last two rules.

    It must have at least two segments (one or more dots).
    Each segment must start with a letter.
    All characters must be alphanumeric or an underscore [a-zA-Z0-9_].


https://developer.android.com/studio/build/configure-app-module#set-application-id

holzeis (changes requested):
This PR breaks the build. The package name needs to be alphanumeric and start with a letter.

```
* What went wrong:
Execution failed for task ':app:processDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > Android resource linking failed
     /Users/holzeis/workspace/coblox/10101/build/app/intermediates/packaged_manifests/debug/AndroidManifest.xml:2: AAPT: error: attribute 'package' in <manifest> tag is not a valid Android package name: 'com.itchysats.10101'.

     /Users/holzeis/workspace/coblox/10101/build/app/intermediates/packaged_manifests/debug/AndroidManifest.xml:22: AAPT: error: attribute 'android:name' in <activity> tag must be a valid Java class name.
```

android/app/build.gradle
klochowicz: it adheres to the last one - the only trouble is the `10101` string. 
I guess this is a question of branding - should we go consistently with `10101` or `TenTenOne` everywhere, or only fallback for the places where we can't use the numbers? @bonomat ?

I will leave this one open and create a subset of this for macos/ios for the demo purposes, as I don't think it breaks those guidelines (I will doublecheck first).

klochowicz: It works well on iOS (the new on the right is after the change)

image

da-kami (approved):

PR #293, opened 2022-11-22
fix: Clamp open channel amount to maker's constraints

klochowicz:
Max allowed channel amount on Lightning in sats: 2^24 - 1 = 16777215. However, we need to
divide it by 3 due to maker's configuration, otherwise it will not be accepted.


klochowicz:
I ended up clamping the amount (initial value and user input) to be within maker's limit.

bonomat (approved):
Easy fix. Thanks

PR #295, opened 2022-11-22
App bar and theme and Dashboard improvements

### LINKED ISSUE #280: Change theme color
# Them color

Change orange to something _softer_. Maybe blue?

_Originally posted by @bonomat in https://github.com/itchysats/10101/discussions/273#discussioncomment-4193804_

One idea is that we could just remove the orange from the top bar because it is too dominant. We could keep the orange in the button (not too dominant on screen).

Another idea is to change the color alltogether, maybe with a pallette that is more like blue/purple/pink, e.g. https://mimo.capital/ (ignore the animations, we no need these)

### LINKED ISSUE #282: Improve the Dashboard
- Have no shadows. Either add or remove shadows from App buttons above.
- Left line of alerts could need some color, maybe orange?
- Make soft corners
- Vertically and horizontally align the icon on the left
- Justify the text
- Change the Bitcoin identifier icon into chain icon

da-kami:
resolves #280
resolves #282

With the later commits the Dashboard changed to:

image

Note: The color can be changed globally with one line if needed :)

### DIFF
--- a/integration_test/wallet_test.dart
+++ b/integration_test/wallet_test.dart
@@ -5,7 +5,6 @@ import 'package:integration_test/integration_test.dart';
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 import 'package:ten_ten_one/balance.dart';
-import 'package:ten_ten_one/wallet/wallet_dashboard.dart';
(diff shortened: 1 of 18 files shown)


bonomat (approved):
Didn't check the code but I like the design!

PR #296, opened 2022-11-22
Align button margins

da-kami:
We are within a `SafeArea` and don't need a margin here, these buttons look different than in other child screens. Removing margin to align.


klochowicz (approved):

da-kami:
bors retry

PR #300, opened 2022-11-22
Showcase more services

da-kami:
This fills up empty space and shows the vision better.

image

image

### DIFF
--- a/lib/models/service_model.dart
+++ b/lib/models/service_model.dart
@@ -1,29 +1,48 @@
 import 'package:flutter/material.dart';
 
-enum ServiceGroup { wallet, trade, bets }
+enum ServiceGroup { wallet, trade, bets, invest }
 
 extension ServiceGroupExtension on ServiceGroup {
   static const labels = {
     ServiceGroup.wallet: "Wallet",
     ServiceGroup.trade: "Trading",
-    ServiceGroup.bets: "Bets"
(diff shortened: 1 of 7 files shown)


lib/models/service_model.dart
```diff
@@ -1,29 +1,48 @@
...
-  static const icons = {Service.cfd: Icons.insights, Service.sportsbet: Icons.sports_football};
+  static const labels = {
+    Service.cfd: "CFD Trading",
+    Service.sportsbet: "Sports Bets",
+    Service.exchange: "Taro Exchange",
+    Service.savings: "Saving"
+  };
+  static const shortLabels = {
+    Service.cfd: "CFDs",
+    Service.sportsbet: "Bets",
```
bonomat, line 34: Nice idea :) 

Is it possible to disable them or show them in a lighter color do indicate they are disabled?

bonomat (approved):

lib/models/service_model.dart
da-kami: How about adding "coming soon" screens instead? - then the button would be enabled but there's only a screen that says coming soon (potentially with some more words on what is planned...?) 
I imagine that could be interesting if anybody joins the demo.

da-kami:
Note: One more commit coming shortly :)

da-kami:
Example placeholder:

image

PR #301, opened 2022-11-22
feat: Add market view

klochowicz:
Embed a TradingView bitmex XBTUSD widget inside a webview.

Note: it works on Android/iOS. Desktop platforms get a placeholder that the
    functionality is not supported yet.

I've recorded in landscape, as it's much nicer for viewing charts like that 😁
https://user-images.githubusercontent.com/8319440/203267520-dd5b8835-12e2-4ca1-8aa9-579b5e8f081a.mov

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -34,7 +34,7 @@ if (keystorePropertiesFile.exists()) {
 }
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion 32
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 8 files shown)


klochowicz:
@holzeis could you please verify that it works on Android too? (I don't have the emulator working...)

holzeis:
> @holzeis could you please verify that it works on Android too? (I don't have the emulator working...)

It unfortunately does not work as is..

> The plugin webview_flutter_android requires Android SDK version 32.

I'll have a look if I can fix that.

holzeis:
> I'll have a look if I can fix that.

This commit 596e0b4551294d27a6917d89f3a69f8055369f26 increases the android sdk versions, required by webview_flutter_android. Only downside we won't be compatible with android versions lower than 19.

holzeis (approved):
https://github.com/flutter/plugins/tree/main/packages/webview_flutter/webview_flutter

The webview_flutter plugin seems to be the far more popular option. No need to action, but leaving it here in case we want to pick it up again after the tournament.

da-kami:
Does this mean we should merge this or not?
It works on iPhone?

klochowicz:
> Does this mean we should merge this or not?
> It works on iPhone?

yes, we should merge it. it works on iPhone - see the video attached.

PR #303, opened 2022-11-22
Update maker spread through api

### LINKED ISSUE #275: Update maker spread through api
In order to mock a profit for the demo, we discussed exposing an api on the maker side to update the spread on an offer. This way we can even go to a negative spread to simulate a profit.

holzeis:
Exposes an api on the maker to set the spread manually. This is for demo purposes!

update spread like this
```
curl -X PUT http://127.0.0.1:8000/api/spread/0 # sets spread to 0.0
curl -X PUT http://127.0.0.1:8000/api/spread/-30 # sets spread to -0.03
```

get the current spread like this
```
curl http://127.0.0.1:8000/api/spread
```

Note, I am dividing any input by 1000 to simplify the put request, also I opted to use an url parameter instead of the body for convenience reasons.

resolves #275

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -3,6 +3,7 @@ use maker::bitmex;
 use maker::cli::Opts;
 use maker::logger;
 use maker::routes;
+use maker::routes::SpreadPrice;
 use std::time::Duration;
 use ten_ten_one::db;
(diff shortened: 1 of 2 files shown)


bonomat (approved):
Very nice.

PR #304, opened 2022-11-22
fix: Use bid / ask price as closing price

holzeis:
The cfd overview was still calculating the pnl with the index price as a closing price.

### DIFF
--- a/lib/cfd_trading/cfd_overview.dart
+++ b/lib/cfd_trading/cfd_overview.dart
@@ -35,7 +35,8 @@ class _CfdOverviewState extends State<CfdOverview> {
     ];
     widgets.addAll(cfds
         .where((cfd) => [CfdState.Open].contains(cfd.state))
-        .map((cfd) => CfdTradeItem(cfd: cfd, closingPrice: offer.index))
+        .map((cfd) => CfdTradeItem(
+            cfd: cfd, closingPrice: cfd.position == Position.Long ? offer.bid : offer.ask))
         .toList());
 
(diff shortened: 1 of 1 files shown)


da-kami (approved):
I think the `bid` and `ask` should be flipped when closing, because we are basically doing a counter position. If we are `Long` we will "go short" to close the position, for going short we would get the selling price, which is the `ask` price.

Bit I might be mistaken - @bonomat ?

luckysori:
> I think the `bid` and `ask` should be flipped when closing, because we are basically doing a counter position. If we are `Long` we will "go short" to close the position, for going short we would get the selling price, which is the `ask` price.
> 
> Bit I might be mistaken - @bonomat ?

Sounds right to me!

bonomat:
> I think the `bid` and `ask` should be flipped when closing, because we are basically doing a counter position. If we are `Long` we will "go short" to close the position, for going short we would get the selling price, which is the `ask` price.

What he did is correct.

```
cfd: cfd, closingPrice: cfd.position == Position.Long ? offer.bid : offer.ask))
```

if your position is `long`, then closing will take the `bid` price. `bid` price is when you go short.

The `askPrice` is always bigger than the `bidPrice`, hence, when you go long, you get the `askPrice` and when you go short you get the `bidPrice`, aka, **buy high, sell low**.

Current prices on BitMex.
```
    "bidPrice": 16195,
    "midPrice": 16195.25,
    "askPrice": 16195.5,
```

bonomat (approved):

PR #307, opened 2022-11-22
Fix: balance is not visible on iphone

### LINKED ISSUE #306: Balance is not visible on iphone 14
Image

holzeis:
Screenshot 2022-11-22 at 13 46 04
Screenshot 2022-11-22 at 13 46 00
Screenshot 2022-11-22 at 13 45 52

The cfd trading screen looks a bit off in comparison to the other screens. But I didn't figure out quickly why this is the case. Leaving it as it is for now, since it isn't a priority for the tournament.

Screenshot 2022-11-22 at 13 51 59

resolves #306

### DIFF
--- a/lib/app_bar_with_balance.dart
+++ b/lib/app_bar_with_balance.dart
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart' hide Divider;
 import 'package:ten_ten_one/utilities/divider.dart';
 
-import 'balance.dart';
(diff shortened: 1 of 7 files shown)


da-kami (approved):
nice thanks! We should always wrap with `SafeArea` - it's interesting that the `Scaffold` does not do that by default. I wonder what would be a usecase where one does not want a `SafeArea` :)

PR #308, opened 2022-11-22
chore: Improve table layout and styling

### LINKED ISSUE #297: Redesign confirmation screen and CFD detail screen
same goes for CFD details screen

image

- the ordering of the details is odd
- expiry shouldn't have a sat sign
- overall it can be a bit _nicer_, e.g. something like ItchySats :D 
- 
image

holzeis:
Screenshot 2022-11-22 at 15 44 17

resolves #297

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -29,7 +29,9 @@ class _CfdOfferState extends State<CfdOffer> {
 
   @override
   Widget build(BuildContext context) {
-    final formatter = NumberFormat.decimalPattern('en');
+    final formatter = NumberFormat();
+    formatter.minimumFractionDigits = 2;
+    formatter.maximumFractionDigits = 2;
 
     final cfdTradingService = context.watch<CfdTradingChangeNotifier>();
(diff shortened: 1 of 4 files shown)


da-kami (approved):
Nice! I like the design, way better!

I would have probably tried to align the items in the second column on the right, but that's not super important

da-kami:
We might want to extract the `info` card that is used for the explanation to the user in other screens, I will see to do this on top of this PR!

bonomat:
making some minor additional adjustments.

bonomat:
> Nice! I like the design, way better!
> 
> I would have probably tried to align the items in the second column on the right, but that's not super important

I'm doing this right now :)

da-kami:
> > Nice! I like the design, way better!
> > I would have probably tried to align the items in the second column on the right, but that's not super important
> 
> I'm doing this right now :)

Great! then I go on testing :)

bonomat:
Minor changes: 

image

image

bonomat (approved):
I'm self-approving my changes :D

bonomat:
bors +

PR #309, opened 2022-11-22
chore: Remove margin from cfd data model

holzeis:
We don't need to store the margin on the cfd as it can be calculated from the other attributes.

Note: this is just a clean up and does not need to get merged. Feel free to skip the PR until after the tournament.

### DIFF
--- a/rust/migrations/20221122111300_drop_margin_from_cfd_table.sql
+++ b/rust/migrations/20221122111300_drop_margin_from_cfd_table.sql
@@ -1,0 +1,3 @@
+-- Add migration script here
+ALTER TABLE
+    cfd DROP COLUMN margin;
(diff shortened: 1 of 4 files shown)


bonomat (approved):
Happy to get this in, but for the record, it is actually the other way around: 
In a perpetual CFD the leverage is calculated from the margin because the margin will increase/decrease over time due to funding rate.

da-kami (commented):
I would not merge this, we had this discussion on itchysats for some time and in the end we regretted not storing the `margin`. I don't think this is needed for the demo :)

da-kami (changes requested):
We regretted not storing the margin in Itchysats in the end. Yes, it can be calculated - but it is a value that does not change - calculation just makes this cumbersome.

bonomat:
Closing this for now as it's not planned.

holzeis:
We should fix how the margin is stored then. Milli sats is unexpected and not consistent with the other properties.

PR #318, opened 2022-11-23
 fix: Mark pending lightning transactions as "off-chain"

klochowicz:
Refactor the switch statement to correctly parse status after data type.

While at there: change the lightning transaction colour slightly to make them
differentiate a bit.

### DIFF
--- a/lib/wallet/payment_history_list_item.dart
+++ b/lib/wallet/payment_history_list_item.dart
@@ -19,46 +19,56 @@ class PaymentHistoryListItem extends StatelessWidget {
   Widget build(BuildContext context) {
     Widget statusIcon;
     var layer = "on-chain";
+    switch (data.type) {
+      case PaymentType.receive:
+        statusIcon = receiveOffChainIcon();
+        layer = "off-chain";
+        break;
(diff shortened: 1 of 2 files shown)


lib/wallet/payment_history_list_item.dart
```diff
@@ -121,6 +131,25 @@ class PaymentHistoryListItem extends StatelessWidget {
...
+          color: Colors.green[700],
+        ));
+  }
+
+  Transform sendOffChainIcon() {
+    return Transform.rotate(
+      angle: 135 * math.pi / 180,
+      child: Icon(
+        FontAwesomeIcons.arrowLeft,
+        color: Colors.red[700],
```
da-kami, line 148: I don't think another shade of `green` / `red` will look good, I think we have enough colors. I'd rather use a different icon, or just use the same color and icon and distinguish by subtitle only.

klochowicz: This is pretty much the same shade, just slightly different tone. I will just use the same colour for now, for the sake of this getting in. Feel free to change to whatever you want later on.

da-kami (approved):

PR #320, opened 2022-11-23
fix: Constrain qr code maximum size

klochowicz:
Just as the receive on-chain, constrain the size of the QR code to 250x250px.
This prevents it from overflowing when width of the app is greater than
height (e.g. in landscape mode on phone or on desktop).


da-kami (approved):
bors r+

PR #321, opened 2022-11-23
It's supposed to be `off-chain`

da-kami:


klochowicz (approved):
thanks for spotting that ;)

PS the commit message is not too great, but I guess you're in a hurry.

da-kami:
> thanks for spotting that ;)
> 
> PS the commit message is not too great, but I guess you're in a hurry.

Yeah, I guess there should always be enough time to write a decent message :) - I'll see to next time again

PR #322, opened 2022-11-23
Add on-/off-chain icons to payment list view

da-kami:


da-kami:
> I like the new icons, they're cool.
> 
> as for the code, it could have been derived from the layer (which should have been an enum in the first place). We can clean this one up later though.

Yeah, I'm generally not in love with that code. We should have created an abstraction similar to `Service` and then put everything we need on there. But I would not bother introducing it now :)

PR #323, opened 2022-11-23
Use a watch channel instead of atomic for SpreadPrice

klochowicz:
Fixes the issue with spread not always being updated properly.

This is more idiomatic way of updating the value from another thread,
e.g. provides more info about what's being updated where.

Proof that it works below: 
```sh
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X GET "http://localhost:8000/api/spread"
0.015%
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X PUT "http://localhost:8000/api/spread"
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X GET "http://localhost:8000/api/offer"                                                                                                                                                                                                         <<<
{"bid":16261.8575,"ask":16843.925,"index":16505.03}%
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X PUT "http://localhost:8000/api/spread/13"
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X GET "http://localhost:8000/api/spread"
0.013%
➜  10101 git:(use-a-watch-channel-instead-of-atomic-state) ✗ curl -X GET "http://localhost:8000/api/offer"
{"bid":16294.8765,"ask":16810.735,"index":16505.03}%
```

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -7,6 +7,7 @@ use maker::routes::SpreadPrice;
 use std::time::Duration;
 use ten_ten_one::db;
 use ten_ten_one::wallet;
+use tokio::sync::watch;
 use tracing::metadata::LevelFilter;
 
 #[rocket::main]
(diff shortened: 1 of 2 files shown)


holzeis:
> Fixes the issue with spread not always being updated properly.

I guess this could as well been fixed by using another ordering strategy. `Ordering::Relaxed` gives you no guarantees - I thought that would be sufficient for this use case. If we wanted more  guarantees - we could have used `Ordering::Aquire`

PR #324, opened 2022-11-23
Handle fields in BitMEX table update separately

luckysori:
After subscribing to the topic `instrument:XBTUSD` on BitMEX, we immediately get the entire table of current values. This meant that we would succeed in initially setting all the `Quote` data.

But every subsequent update that we get via the websocket is not guaranteed to contain all of `bidPrice`, `askPrice` and `markPrice`. Consequently we were almost always failing to deserialise the table updates, because all of those fields were mandatory.


da-kami (approved):
LGTM

PR #334, opened 2022-11-23
fix: Recover connection after taker or maker disconnect.

### LINKED ISSUE #341: Recover connections after maker or taker has been disconnected
Regardless if maker, taker or both are offline the app should be able to recover from any situation given that both parties are back online again.

In particular this relates to the following error.
```
flutter: {CfdOrderConfirmation} {build} {Failed to open CFD: Peer for first hop currently disconnected/pending monitor update!} {FfiException(RESULT\_ERROR, Peer for first hop currently disconnected/pending monitor update!, null)} {LogLevel.ERROR} {21 November 2022 02:04:37 PM}
```

holzeis:
This PR improves the way 10101 is managing connections with the maker by introducing a dedicated connect api.

Among other improvements this PR fixes the following error.

```
flutter: {CfdOrderConfirmation} {build} {Failed to open CFD: Peer for first hop currently disconnected/pending monitor update!} {FfiException(RESULT\_ERROR, Peer for first hop currently disconnected/pending monitor update!, null)} {LogLevel.ERROR} {21 November 2022 02:04:37 PM}
```

 - Instead of blocking indefinitely when creating a channel to keep the channel alive, we now connect to the maker upon startup. Before a new thread was kept alive indefinitely everytime a new channel has been created. The thread pool was very limited though resulting into other request not getting processed and blocked!
 
 - Check if the maker is alive and try to connect once alive. Before we never checked if the connection actually succeeded, I've added a corresponding error handling. I am not very happy with checking via an http api if the maker is still alive. We see in the logs that the peer manager gets notified about a disconnect, but I didn't find a way of getting that out of the peer manager - hence the workaround over the maker alive api.

connect

An additional improvement would be to have only one thread indefinitely running and combine the run ldk node and connect together into one api call. I will create a follow up PR for that, but this is already sufficient for a testnet release.

resolves #341

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -208,6 +208,11 @@ class _TenTenOneState extends State<TenTenOneApp> {
           .then((value) => FLog.info(text: "ldk node stopped."))
           .catchError((error) => FLog.error(text: "ldk stopped with an error", exception: error));
 
+      // connect to the maker, this will not return unless there was an error.
+      api.connect().then((value) => FLog.error(text: "Lost connection to the maker")).catchError(
+          (error) =>
+              FLog.error(text: "Lost connection to the maker with an error", exception: error));
+
       FLog.info(text: "TenTenOne is ready!");
       setState(() {
         ready = true;
--- a/lib/wallet/open_channel.dart
+++ b/lib/wallet/open_channel.dart
@@ -1,6 +1,7 @@
 import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart' hide Divider;
 import 'package:flutter/services.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 import 'package:ten_ten_one/balance.dart';
@@ -121,12 +122,19 @@ class _OpenChannelState extends State<OpenChannel> {
                   onPressed: () async {
                     FLog.info(
                         text: "Opening Channel with capacity " + takerChannelAmount.toString());
-
-                    api.openChannel(takerAmount: takerChannelAmount);
-                    ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
-                      content: Text("Waiting for channel to get established"),
-                    ));
-                    context.go('/');
+                    try {
+                      await api.openChannel(takerAmount: takerChannelAmount);
+                      ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
+                        content: Text("Waiting for channel to get established"),
+                      ));
+                      context.go('/');
+                    } on FfiException catch (error) {
+                      FLog.error(text: "Failed to open channel.", exception: error);
+                      ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
+                        backgroundColor: Colors.red,
+                        content: Text("Failed to open channel. Is the maker online?"),
+                      ));
+                    }
                   },
                   child: const Text('Open Channel')),
             ),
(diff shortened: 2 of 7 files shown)


da-kami (approved):
I'm happy to see this merge, but for the future (i.e. after tournament) I think we should re-think the pattern of "always keeping the connection alive" between taker and maker. With this pattern we rely on the connection management, because each action that triggers a call to the maker expects the connection to be alive.
It may look like this is more stable, but when running the application for longer time we might encounter problems that are hard to analyze - that's what we encountered in ItchySats.

Problems that we encountered with this pattern in itchysats were, that the application thought the connection was alive (because no timeout was reached, no error reported), but we were unable to send/receive over the connection. Additionally, keeping the connection alive indefinitely puts unnecessary load on the maker. If we have some takers that just keep idling without doing anything we still keep their connection alive as long as the application is running.
 
I wonder if it is feasible to just establish a connection when needed; i.e. call `connect_peer_if_necessary` automatically before a call to the maker. Can we not rely on `connect_peer_if_necessary` establishing a connection when needed?

lib/main.dart
```diff
@@ -208,6 +208,11 @@ class _TenTenOneState extends State<TenTenOneApp> {
           .then((value) => FLog.info(text: "ldk node stopped."))
           .catchError((error) => FLog.error(text: "ldk stopped with an error", exception: error));
 
+      // connect to the maker, this will not return unless there was an error.
+      api.connect().then((value) => FLog.error(text: "Lost connection to the maker")).catchError(
+          (error) =>
+              FLog.error(text: "Lost connection to the maker with an error", exception: error));
```
luckysori, line 214: ❓ Just for my understanding: The comment says that this shouldn't return unless there is an error, but there is more code that needs to run after this. Does this work because we are not `await`ing this future?

luckysori (approved):
Thank you for fixing this!

klochowicz:
> I'm happy to see this merge, but for the future (i.e. after tournament) I think we should re-think the pattern of "always keeping the connection alive" between taker and maker. With this pattern we rely on the connection management, because each action that triggers a call to the maker expects the connection to be alive. It may look like this is more stable, but when running the application for longer time we might encounter problems that are hard to analyze - that's what we encountered in ItchySats.
> 
> Problems that we encountered with this pattern in itchysats were, that the application thought the connection was alive (because no timeout was reached, no error reported), but we were unable to send/receive over the connection. Additionally, keeping the connection alive indefinitely puts unnecessary load on the maker. If we have some takers that just keep idling without doing anything we still keep their connection alive as long as the application is running.
> 
> I wonder if it is feasible to just establish a connection when needed; i.e. call `connect_peer_if_necessary` automatically before a call to the maker. Can we not rely on `connect_peer_if_necessary` establishing a connection when needed?

note: it will be hard to say when it is necessary. if we keep going with the model that our maker is routing the payments then of course we need to keep it open at all times, otherwise they can't get their lightning invoices paid etc. (it's important for the lightning wallet functionality, not the cfd trading)

klochowicz (approved):

luckysori:
> note: it will be hard to say when it is necessary. if we keep going with the model that our maker is routing the payments then of course we need to keep it open at all times, otherwise they can't get their lightning invoices paid etc. (it's important for the lightning wallet functionality, not the cfd trading)

What if you just reestablish the connection when you need to make a payment?

I'm not sure what the expectation is for Lightning nodes in general. Perhaps you are meant to keep the connection alive with all your channel peers. Although the channel obviously remains open and safe if you lose the connection with the counterparty.

lib/main.dart
holzeis: yes.. with`then` and `catchError` we register callback functions in case the function returns.. we do the same with runLdk.

Issue #339, opened 2022-11-23
Division by zero


holzeis:
In case the maker is offline the app crashes with a division by zero error on the cfd_overview and cfd_order_details view.

Precondition: open cfd

![Image](https://user-images.githubusercontent.com/382048/203627462-13014d93-e732-4216-87d8-e07bacc46e01.png)

da-kami:
Interesting, I thought that was resolved with #327 - seems we have even more of these.

PR #343, opened 2022-11-23
fix: Prevent division by zero

klochowicz:
This fixes the problem when we had opening price, but no closing price.


luckysori (approved):
Haha, I knew I was gonna make a mistake. Thanks for fixing it :D

da-kami:
If this fixes https://github.com/itchysats/10101/issues/339 we should close the ticket with this PR ;)

PR #349, opened 2022-11-24
A bunch of improvements for ldk events handling

klochowicz:
- don't use a Filter (no need)
- handle missing events
- reduce log spam


bonomat (approved):
yay, less log spam :)

PR #358, opened 2022-11-24
fix: Catch exception during wallet sync

klochowicz:
All anyhow::Result<> going through API are coming up as exceptions, and we need
to be mindful whenever we call Rust api that returns one - we should always
wrap them in a try/catch block.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -230,10 +230,14 @@ class _TenTenOneState extends State<TenTenOneApp> {
   }
 
   Future<void> _callSync() async {
-    final balance = await api.getBalance();
-    bitcoinBalance.update(Amount(balance.onChain.confirmed));
-    lightningBalance.update(Amount(balance.offChain));
-    FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    try {
+      final balance = await api.getBalance();
+      bitcoinBalance.update(Amount(balance.onChain.confirmed));
+      lightningBalance.update(Amount(balance.offChain));
+      FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    } catch (error) {
+      FLog.error(text: "Failed to sync wallet:" + error.toString());
+    }
   }
 
   Future<void> _callGetOffers() async {


lib/main.dart
```diff
@@ -230,10 +230,14 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
-    bitcoinBalance.update(Amount(balance.onChain.confirmed));
-    lightningBalance.update(Amount(balance.offChain));
-    FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    try {
+      final balance = await api.getBalance();
+      bitcoinBalance.update(Amount(balance.onChain.confirmed));
+      lightningBalance.update(Amount(balance.offChain));
+      FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    } catch (error) {
+      FLog.error(text: "Failed to initialise: Unknown error");
```
holzeis, line 239: would be nice to print the error here as well. you can add the error with the exception parameter.

lib/main.dart
```diff
@@ -230,10 +230,14 @@ class _TenTenOneState extends State<TenTenOneApp> {
   }
 
   Future<void> _callSync() async {
-    final balance = await api.getBalance();
-    bitcoinBalance.update(Amount(balance.onChain.confirmed));
-    lightningBalance.update(Amount(balance.offChain));
-    FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    try {
```
holzeis, line 233: 🤔 run periodically is already wrapping that into a try catch - why is this needed?

lib/main.dart
```diff
@@ -230,10 +230,14 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
+      final balance = await api.getBalance();
+      bitcoinBalance.update(Amount(balance.onChain.confirmed));
+      lightningBalance.update(Amount(balance.offChain));
+      FLog.trace(text: 'Successfully synced Bitcoin wallet');
+    } catch (error) {
+      FLog.error(text: "Failed to sync wallet:" + error.toString());
```
holzeis, line 239: FLog.error takes an exception argument. I think its nicer to put the error there, instead of adding it to the text.

PR #361, opened 2022-11-24
Fix value refresh on price change

da-kami:
Note: @luckysori also fixed this on `fix/responsive-offer-lucas` - if the solution is equivalent we can close this PR as well :)


holzeis (approved):
Thanks!

luckysori (approved):
LGTM. Can't really argue that my extremely similar [solution](https://github.com/itchysats/10101/commit/9317bf84f5c037850a751fa3001f7f6302ef5708) is any better!

Issue #336, opened 2022-11-23
Add testnet configuration for released app


holzeis:
In case the app is released it should point to the testnet maker, etc. instead of regtest.

klochowicz:
it seems it should be enough to swap the `maker_pk()` to hard-coded testnet key in the release mode?

holzeis:
> it might be enough to swap the `maker_pk()` to hard-coded testnet key in the release mode?

Not just the public key, but also the endpoint and the network. - network is currently provided through flutter though. For consistency this should probably also be configured in `wallet.rs` (still not a fan of managing configuration in code 🙈 )

PR #363, opened 2022-11-24
feat: Support testnet

klochowicz:
Revamp configuration to pass NETWORK env variable from command line (both for
taker and maker).
In case it's not set, or it's invalid, it defaults to testnet.

Move all the config values to a dedicated module to clear the code.

For local development, I recommend to set `export NETWORK=regtest` in your
bashrc to preserve previous behaviour.


holzeis (approved):
Unsure if we really have to adapt the maker as well, but this will do for the 0.2.0 release.

holzeis:
resolves #336

PR #364, opened 2022-11-24
Feat/hide open channel card on creation

### LINKED ISSUE #337: Hide open channel card if channel is in creation
The open channel card should be gone after successfully initiating a channel. It might reappear if eventually the channel creation failed, but keeping it in the meantime will confuse the user and may break things if the user retries to open the channel.

An easy quick fix could be a simple timeout, which waits for a lightning balance. If within the timeout nothing changed the card could reappear again.

holzeis:
open-channel

resolves #337


luckysori (approved):
LGTM.

Didn't thoroughly review the code, but we can always go back and fix things if we need to.

PR #365, opened 2022-11-24
feat: In app splash screen

holzeis:
Not an native splash screen, but at least the screen is not black until the app is ready.

startup


luckysori (approved):
LGTM!

PR #376, opened 2022-11-25
Log `Maker is alive` on trace

da-kami:
This is super spammy in the logs so I think we should have it on `trace`. If needed the taker can be run on `trace` to see the log.


klochowicz:
this is part of another PR that I have opened already 😂

luckysori (approved):

PR #377, opened 2022-11-25
Fix open CFD functionality

### LINKED ISSUE #369: Could not retrieve short channel id
The error is reproducible by the following steps

1. Fund on-chain wallet
2. Open channel
3. faucet
4. Open cfd

```
{Failed to open CFD: Could not retrieve short channel id} {FfiException(RESULT_ERROR, Could not retrieve short channel id, null)}
```

luckysori:
Fixes #369.

Interestingly, reverting https://github.com/itchysats/10101/commit/38c09b25c8903e2aeb6831c637493f88e5731ff1 does the trick.

My theory is that the taker wasn't able to pick up the funding transaction because of the missing filter, so the channel would never be considered ready and the short channel ID would never be available when opening a CFD.

### DIFF
--- a/rust/src/config.rs
+++ b/rust/src/config.rs
@@ -9,9 +9,6 @@ const MAINNET_ELECTRUM: &str = "ssl://blockstream.info:700";
 const TESTNET_ELECTRUM: &str = "ssl://blockstream.info:993";
 const REGTEST_ELECTRUM: &str = "tcp://localhost:50000";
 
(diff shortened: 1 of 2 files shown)


klochowicz:
oh wow, sorry for introducing such a regression!

klochowicz (approved):

bonomat (approved):
Awesome finding :)

luckysori:
> oh wow, sorry for introducing such a regression!

No hard feelings! Could have happened to anyone (and we clearly approved it!).

PR #379, opened 2022-11-25
chore: Set Makefile to run on regtest by default

klochowicz:
Makefile is for devs, so let's make it more convenient.

note: plain `flutter run` still will require a `NETWORK` override

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,9 @@ To run on `regtest` you have to change the default network (`testnet`) to `regte
 
 Example command for maker:
 
-`NETWORK=regtest make maker`
+`make maker`
+
+(`make` runs prepend NETWORK=regtest automatically)
 
 Example command for taker:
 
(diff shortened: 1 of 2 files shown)


da-kami (approved):
Couldn't we change the readme for the taker command to `make run` then?
I actually never used that, but might be more intuitive :)

README.md
```diff
@@ -86,7 +86,9 @@ To run on `regtest` you have to change the default network (`testnet`) to `regte
 
 Example command for maker:
 
-`NETWORK=regtest make maker`
+`make maker`
+
+(`make` runs prepend NETWORK=regtest automatically)
 
 Example command for taker:
```
da-kami, line 93: Maybe change the command to `make run`? 
(above as well...)

holzeis (approved):

README.md
holzeis: like `make run maker`, `make run taker`?

klochowicz: the commands I changed are not just for running, they build too, hence I'm hesitant to change to *run*.

PR #386, opened 2022-11-26
fix: Skip syncing the wallet when sending to address

### LINKED ISSUE #388: Operation timeout when opening a channel
env: testnet

Opening a channel on testnet runs into a timeout when calling the open channel api on the maker to send funds to the taker. The fund eventually end up at the taker but, since the request times out the taker already returns an error.

Screenshot_2022_1126_102104

holzeis:
Send to address is called from the open channel route and takes quite some time - letting the client timeout for the request and the channel fails to open.

Sync is already running every 10 seconds - so the wallet should anyways be up to date.

Hard to test since it's not reproducible locally, but I think this resolves #388

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -248,8 +248,6 @@ impl Wallet {
     }
 
     pub fn send_to_address(&self, send_to: Address, amount: u64) -> Result<Txid> {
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Agreed

PR #387, opened 2022-11-26
chore: Combine run ldk and connect into a single thread

holzeis:
Both actions are blocking, this will free up a thread.

### DIFF
--- a/rust/src/connection.rs
+++ b/rust/src/connection.rs
@@ -1,0 +1,56 @@
+use crate::config;
+use crate::wallet;
+use anyhow::Result;
+
+/// blocks the current thread, if you want this to be run asynchronously you need to spawn a thread
+/// from the outside.
+pub async fn keep_alive() -> Result<()> {
+    let mut connected = false;
+    loop {
+        let alive = check_alive().await?;
+
+        if !connected || !alive {
+            connected = connect().await?;
+        }
+
+        // looping here indefinitely to keep the connection with the maker alive.
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+    }
+}
+
+async fn check_alive() -> Result<bool> {
+    tracing::trace!("Checking if maker is alive");
+    let client = reqwest::Client::builder()
+        .timeout(config::TCP_TIMEOUT)
+        .build()?;
+    let result = client
+        .get(config::maker_endpoint() + "/api/alive")
+        .send()
+        .await;
+
+    let alive = match result {
+        Ok(_) => true,
+        Err(err) => {
+            tracing::warn!("Maker is offline! {err:?}");
+            false
+        }
+    };
+
+    Ok(alive)
+}
+
+async fn connect() -> Result<bool> {
+    let result = wallet::connect().await;
+    let connected = match result {
+        Ok(()) => {
+            tracing::info!("Successfully connected to maker");
+            true
+        }
+        Err(err) => {
+            tracing::warn!("Failed to connect to maker! {err:?}");
+            false
+        }
+    };
+
+    Ok(connected)
+}
--- a/rust/src/lightning.rs
+++ b/rust/src/lightning.rs
@@ -17,7 +17,6 @@ use bdk::blockchain::ElectrumBlockchain;
 use bdk::database::MemoryDatabase;
 use bdk::wallet::time::get_timestamp;
 use bitcoin_bech32::WitnessProgram;
-use futures::executor::block_on;
 use lightning::chain;
 use lightning::chain::chaininterface::BroadcasterInterface;
 use lightning::chain::chaininterface::ConfirmationTarget;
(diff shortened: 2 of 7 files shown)


rust/src/connection.rs
```diff
@@ -0,0 +1,56 @@
...
+use anyhow::Result;
+
+/// blocks the current thread, if you want this to be run asynchronously you need to spawn a thread
+/// from the outside.
+pub async fn keep_alive() -> Result<()> {
+    let mut connected = false;
+    loop {
+        let alive = check_alive().await?;
+
+        if !connected || !alive {
```
bonomat, line 12: I'm not sure this makes 100% sense, help me out please :) 

- if `alive=false` the maker's http is not reachable but we try to connect via LDK 
- if `alive=true` the maker's http is reachable (which should always be the case), but we could have lost the connection the his LDK and will never reconnect

bonomat (commented):
I'm not sure about the `alive` logic

rust/src/connection.rs
```diff
@@ -0,0 +1,56 @@
...
+        // looping here indefinitely to keep the connection with the maker alive.
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+    }
+}
+
+async fn check_alive() -> Result<bool> {
+    tracing::trace!("Checking if maker is alive");
+    let client = reqwest::Client::builder()
+        .timeout(config::TCP_TIMEOUT)
+        .build()?;
```
luckysori, line 25: 🔧 Would it be safe to `expect` here? It would be nice if `check_alive` were infallible.

rust/src/connection.rs
```diff
@@ -0,0 +1,56 @@
...
+        Err(err) => {
+            tracing::warn!("Maker is offline! {err:?}");
+            false
+        }
+    };
+
+    Ok(alive)
+}
+
+async fn connect() -> Result<bool> {
```
luckysori, line 42: 🙃 Maybe a bit more idiomatic to return a `Result<()>` here.

It would simplify the code and above you could do:

```rust
connected = connect().await.is_ok();
```

---

As a side note, currently this function is infallible, so you don't need the `Result` in it's current form!

rust/src/connection.rs
luckysori: It's the same logic as before this change though. 

I agree that we should find a way to notice that the LDK connection is dead, but perhaps in a separate PR?

luckysori (approved):
LGTM!

rust/src/lightning.rs
```diff
@@ -587,22 +577,6 @@ pub async fn run_ldk(system: &LightningSystem) -> Result<BackgroundProcessor> {
         Some(scorer),
     );
 
-    let peer_data_path = format!("{ldk_data_dir}/channel_peer_data");
-    let mut info = disk::read_channel_peer_data(Path::new(&peer_data_path))?;
-    for (pubkey, peer_addr) in info.drain() {
-        for chan_info in system.channel_manager.list_channels() {
-            if pubkey == chan_info.counterparty.node_id {
-                block_on(async {
-                    let _ = connect_peer_if_necessary(
```
da-kami, line 596: ❓ The reconnect is handled automatically based on the `channel_peer_data` file? Is this file persisted forever?

I'm a bit confused by the commit message

> I noticed that after combining the connection and run ldk functions, that the connection never got re-established after restarting. 

So, we do reconnect automatically after removing this code?

da-kami (approved):

rust/src/connection.rs
klochowicz: > 🔧 Would it be safe to `expect` here? It would be nice if `check_alive` were infallible.

it should be, we're passing static arguments after all into the builder.

rust/src/connection.rs
holzeis: @bonomat you are completely right. like mentioned in the last PR, it would be much nicer if we handle the actual event instead of checking if the maker is alive. This was a workaround - this PR is only a small clean up to not block another thread. I am creating a ticket for the handling the actual event - https://github.com/itchysats/10101/issues/433

rust/src/lightning.rs
holzeis: > ❓ The reconnect is handled automatically based on the `channel_peer_data` file? Is this file persisted forever?

yes, it has been written to the file system, however we don't need that, for the reasons described in the commit message.

> So, we do reconnect automatically after removing this code?

yes.

rust/src/connection.rs
holzeis: thanks for the feedback, I adapted the implementation, but figured the improvements were not worthwhile the additional round of testing. Also giving that this is very much still a workaround patch.

rust/src/connection.rs
holzeis: 💯 thanks for the feedback, I played around with it a bit, but opted against it to not risk introducing some regression.

PR #392, opened 2022-11-28
sync-optimizations

bonomat:
What do you guys think?

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use bdk::bitcoin::Network;
 use maker::bitmex;
 use maker::cli::Opts;
 use maker::logger;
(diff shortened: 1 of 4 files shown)


maker/src/main.rs
```diff
@@ -44,11 +46,23 @@ async fn main() -> Result<()> {
...
+            let duration = started.elapsed();
+            let sync_time_in_seconds = duration.as_secs();
+            match wallet_result {
+                Ok(balance) => tracing::info!(?balance, sync_time_in_seconds, "Current balance"),
+                Err(e) => {
+                    tracing::error!(sync_time_in_seconds, "Could not retrieve balance: {e:#}")
+                }
             }
-            tokio::time::sleep(Duration::from_secs(10)).await;
+            let sync_time = match network {
```
da-kami, line 59: Am I understanding this correct that the  `maker` has a less frequent sync interval than the `taker`? 
I wonder if that can cause errors where the taker has already caught up with a state on chain, whereas the maker did not and errors.

da-kami (approved):
Detach fetching the balance from the sync is a good idea!
This might affect the UXa, because if we sync less frequent the user might have to wait for long to see a balance change after e.g. initial deposit of BTC. I am not sure we can improve that, because there is no good indicator for triggering a sync for the initial deposit (because the trigger is a transaction that is outside of the application). Idea for the future: We could temporarily set a higher sync interval for `x` time if the user navigated to the `Receive on-chain` screen.

---

We could go one step further and let the wallet handle the sync interval internally in the rust code instead of triggering it from the outside, but not sure if that will make it more stable.

maker/src/main.rs
bonomat: It's time to find out 😅.

bonomat:
> We could go one step further and let the wallet handle the sync interval internally in the rust code instead of triggering it from the outside, but not sure if that will make it more stable.

I thought about it but decided against it for now. Maybe something we want to keep in mind for `v1`

PR #394, opened 2022-11-28
Allow taker to close channel with 10101 maker

### LINKED ISSUE #373: Add `close channel ` action to the Lightning wallet actions in the UI
As mentioned in https://github.com/itchysats/10101/issues/372 the user should have the means to back out of the channel if something goes wrong. 

The easiest way to achieve this is adding another action to the FAB in the Lightning wallet that allows the user to close the channel. 
We *could* restrict this to only allow it if there is no CFD currently open, to keep the scope smaller.

luckysori:
Fixes #373.

For now it defaults to a collaborative closure. Eventually, we might want to let the taker force-close.

https://user-images.githubusercontent.com/9418575/204174964-f09e3854-f911-496a-abfb-24b4a9962530.mp4

### DIFF
--- a/lib/wallet/close_channel.dart
+++ b/lib/wallet/close_channel.dart
@@ -1,0 +1,109 @@
+import 'package:f_logs/f_logs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
+import 'package:go_router/go_router.dart';
+import 'package:provider/provider.dart';
+import 'package:ten_ten_one/cfd_trading/validation_error.dart';
+import 'package:ten_ten_one/ffi.io.dart';
+import 'package:ten_ten_one/models/amount.model.dart';
+import 'package:ten_ten_one/models/balance_model.dart';
+import 'package:ten_ten_one/utilities/submit_button.dart';
+
+class CloseChannel extends StatefulWidget {
+  const CloseChannel({Key? key}) : super(key: key);
+
+  static const route = '/' + subRouteName;
+  static const subRouteName = 'close-channel';
+
+  @override
+  State<CloseChannel> createState() => _CloseChannelState();
+}
+
+class _CloseChannelState extends State<CloseChannel> {
+  late AmountDisplay closeAmount;
+
+  @override
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    LightningBalance lightningBalance = context.watch<LightningBalance>();
+    closeAmount = lightningBalance.amount.display(currency: Currency.sat);
+
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('Close channel'),
+      ),
+      body: SafeArea(
+          child: ListView(
+        children: [
+          Container(
+            padding: const EdgeInsets.all(20.0),
+            child: Column(children: [
+              Center(
+                child: RichText(
+                    text: TextSpan(
+                        style: const TextStyle(color: Colors.black, fontSize: 18),
+                        children: [
+                      const TextSpan(
+                          text: "This will close your Lightning channel with the 10101 maker\n\n",
+                          style: TextStyle(fontWeight: FontWeight.bold)),
+                      TextSpan(
+                          text:
+                              "Closing the channel will send ${closeAmount.value} sats to your 10101 on-chain wallet.\n\n"),
+                      const TextSpan(
+                          text:
+                              "Once the channel is closed, you will have to open a new one if you want to continue using your wallet for Lightning payments and trading.")
+                    ])),
+              ),
+            ]),
+          ),
+          Align(
+              alignment: Alignment.bottomCenter,
+              child: Container(
+                  padding: const EdgeInsets.all(20.0),
+                  child: Column(
+                    mainAxisAlignment: MainAxisAlignment.end,
+                    children: [
+                      AlertMessage(
+                          message: Message(
+                              title:
+                                  "Every time you open a new channel, you incur in additional transaction fees.",
+                              type: AlertType.info)),
+                      const SizedBox(height: 50),
+                    ],
+                  ))),
+          Padding(
+              padding: const EdgeInsets.only(right: 20.0),
+              child: Container(
+                alignment: Alignment.bottomRight,
+                child: SubmitButton(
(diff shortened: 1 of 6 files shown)


lib/wallet/close_channel.dart
```diff
@@ -0,0 +1,118 @@
+import 'package:f_logs/f_logs.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
+import 'package:go_router/go_router.dart';
+import 'package:provider/provider.dart';
+import 'package:ten_ten_one/cfd_trading/validation_error.dart';
+import 'package:ten_ten_one/ffi.io.dart';
+import 'package:ten_ten_one/models/amount.model.dart';
+import 'package:ten_ten_one/models/balance_model.dart';
+import 'package:ten_ten_one/utilities/submitting_buton_icon.dart';
```
da-kami, line 10:
```suggestion
import 'package:ten_ten_one/utilities/submitting_button_icon.dart';
```

da-kami (commented):
I think this action should only be available if there is not open CFD. Unless I missed it the user can always trigger it at the moment.
If a CFD is open the channel `closeAmount` will be incorrect. 

We should be able to tell if a CFD is open based on `cfds` of the `CfdTradingChangeNotifier`. I would still display the action in the list, but deactivate the `Confirm` button with an info box if a CFD is open.

luckysori:
> I think this action should only be available if there is not open CFD. Unless I missed it the user can always trigger it at the moment. If a CFD is open the channel `closeAmount` will be incorrect.
> 
> We should be able to tell if a CFD is open based on `cfds` of the `CfdTradingChangeNotifier`. I would still display the action in the list, but deactivate the `Confirm` button with an info box if a CFD is open.

Agreed, but the speed dial is already hidden if we don't have any CFDs open:

https://github.com/itchysats/10101/blob/b31dac088dd4500986ef9a19fb6422dd0af9b8d3/lib/wallet/wallet_lightning.dart#L80

We've just added another action that cannot be triggered if a CFD is open.

da-kami:
> > I think this action should only be available if there is not open CFD. Unless I missed it the user can always trigger it at the moment. If a CFD is open the channel `closeAmount` will be incorrect.
> > We should be able to tell if a CFD is open based on `cfds` of the `CfdTradingChangeNotifier`. I would still display the action in the list, but deactivate the `Confirm` button with an info box if a CFD is open.
> 
> Agreed, but the speed dial is already hidden if we don't have any CFDs open:
> 
> https://github.com/itchysats/10101/blob/b31dac088dd4500986ef9a19fb6422dd0af9b8d3/lib/wallet/wallet_lightning.dart#L80
> 
> We've just added another action to the speed dial that cannot be triggered if a CFD is open.

Fair enough - might be bit confusing to the user that the action is only shown if there are no CFDs. It's hard to understand for a user 😅

da-kami:
https://github.com/itchysats/10101/pull/398 is relevant for this PR, depending on what is merged first we will have to rebase one on the other :)

luckysori:
> Fair enough - might be bit confusing to the user that the action is only shown if there are no CFDs. It's hard to understand for a user 😅

That is why we have this message!

https://github.com/itchysats/10101/blob/b31dac088dd4500986ef9a19fb6422dd0af9b8d3/lib/wallet/wallet_lightning.dart#L56-L64

lib/utilities/submitting_button_icon.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:flutter/material.dart';
+
+class SubmittingButtonIcon extends StatelessWidget {
```
holzeis, line 3: https://github.com/itchysats/10101/pull/398 is introducing an AsyncButton, which we should use eventually.

lib/wallet/close_channel.dart
```diff
@@ -0,0 +1,118 @@
...
+        ],
+      )),
+      floatingActionButton: Row(
+        mainAxisAlignment: MainAxisAlignment.end,
+        children: [
+          Padding(
+              padding: const EdgeInsets.only(right: 20.0),
+              child: ElevatedButton.icon(
+                label: const Text('Confirm'),
+                icon: !submitting ? const SizedBox.shrink() : const SubmittingButtonIcon(),
```
holzeis, line 88: ❓ is there a reason you are using SizedBox.shrink() instead of an empty Container? I think a container would have 0 width and 0 height by default?

lib/wallet/close_channel.dart
```diff
@@ -0,0 +1,118 @@
...
+    setState(() {
+      submitting = true;
+    });
+
+    FLog.info(text: "Closing channel with outbound liquidity of " + closeAmount.toString());
+
+    try {
+      await api.closeChannel();
+    } on FfiException catch (error) {
+      submitting = false;
```
holzeis, line 106: 🤔 does that work if we are not setting the state?

holzeis (approved):
LGTM 👍 A screenshot would have been nice 😅

luckysori:
> LGTM 👍 A screenshot would have been nice 😅

But I posted a whole video!

lib/wallet/close_channel.dart
luckysori: I tried using a `Container` at first, but it would cause the button to have the height of the whole screen.

lib/wallet/close_channel.dart
luckysori: Good catch! I copied it wrong from the open-channel code.

da-kami:
@luckysori I just had a look, but I don't think this is rebased to use `SubmitButton` yet - was something not pushed?

luckysori:
> @luckysori I see that you pinged me here and I just had a look, but I don't think this is rebased to use `SubmitButton` yet - was something not pushed?

I pinged you a long time ago, so we can consider that ping stale :D

lib/wallet/close_channel.dart
holzeis: oh my, was that already there yesterday? 😅  I somehow overlooked that 🙈

holzeis:
> > LGTM 👍 A screenshot would have been nice 😅
> 
> But I posted a whole video!

oh my, was that already there yesterday? 😅 I somehow overlooked that 🙈

PR #395, opened 2022-11-28
Add sqlite support for storing lightning payments

klochowicz:
note: this is just the insert/load API with tests, adding it now to avoid trouble rebasing later (and to spark a conversation whether we should use one or two tables (inbound / outbound payments)

### DIFF
--- a/rust/migrations/20221128012310_add_payments.sql
+++ b/rust/migrations/20221128012310_add_payments.sql
@@ -1,0 +1,12 @@
+-- Add lightning payments table
+CREATE TABLE IF NOT EXISTS payments (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    payment_hash TEXT UNIQUE NOT NULL,
+    preimage TEXT,
+    secret TEXT,
+    htlc_status TEXT NOT NULL,
+    amount_msat INTEGER,
+    created INTEGER NOT NULL,
+    updated INTEGER NOT NULL
+);
+CREATE UNIQUE INDEX IF NOT EXISTS payment_hash ON payments (payment_hash);
--- a/rust/migrations/20221128012310_add_payments.sql
+++ b/rust/migrations/20221128012310_add_payments.sql
@@ -4,6 +4,7 @@ CREATE TABLE IF NOT EXISTS payments (
     payment_hash TEXT UNIQUE NOT NULL,
     preimage TEXT,
     secret TEXT,
+    flow TEXT NOT NULL,
     htlc_status TEXT NOT NULL,
     amount_msat INTEGER,
     created INTEGER NOT NULL,
--- a/rust/src/db.rs
+++ b/rust/src/db.rs
@@ -126,6 +126,7 @@ pub async fn insert_payment(payment: &PaymentInfo) -> Result<()> {
         hash,
         preimage,
         secret,
+        flow,
         status,
         amt_msat,
         created_timestamp,
@@ -134,15 +135,16 @@ pub async fn insert_payment(payment: &PaymentInfo) -> Result<()> {
 
     let query_result = sqlx::query(
         r#"
-        INSERT INTO payments (payment_hash, preimage, secret, htlc_status, amount_msat, created, updated)
-        VALUES ($1, $2, $3, $4, $5, $6, $7)
+        INSERT INTO payments (payment_hash, preimage, secret, flow, htlc_status, amount_msat, created, updated)
+        VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
         "#,
     )
     .bind(base64::encode(hash.0))
     .bind(preimage.map(|x| base64::encode(x.0)))
     .bind(secret.map(|x| base64::encode(x.0)))
+    .bind(flow)
     .bind(status)
-        .bind(amt_msat.0.map(|msat| msat as i64))
+    .bind(amt_msat.0.map(|msat| msat as i64))
     .bind(created_timestamp as i64)
     .bind(updated_timestamp as i64)
     .execute(&mut conn)
@@ -201,6 +203,7 @@ pub async fn load_payments() -> Result<Vec<PaymentInfo>> {
                 payment_hash,
                 preimage,
                 secret,
+                flow as "flow: crate::lightning::Flow",
                 htlc_status as "status: crate::lightning::HTLCStatus",
                 amount_msat,
                 updated,
@@ -222,6 +225,7 @@ pub async fn load_payments() -> Result<Vec<PaymentInfo>> {
             secret: row
                 .secret
                 .map(|x| PaymentSecret(into_array(base64::decode(x).unwrap()))),
+            flow: row.flow,
             status: row.status,
             amt_msat: row
                 .amount_msat
@@ -238,6 +242,7 @@ pub async fn load_payments() -> Result<Vec<PaymentInfo>> {
 
 #[cfg(test)]
 mod tests {
+    use crate::lightning::Flow;
     use crate::lightning::MillisatAmount;
     use bdk::wallet::time::get_timestamp;
     use std::env::temp_dir;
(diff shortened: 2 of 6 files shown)


klochowicz:
converting to draft as we still obviously need payment hash😓

rust/src/lightning.rs
```diff
@@ -974,7 +975,7 @@ async fn handle_ldk_events(
                     Entry::Occupied(mut e) => {
                         let payment = e.get_mut();
                         payment.status = HTLCStatus::Succeeded;
-                        payment.timestamp = get_timestamp();
+                        payment.updated_timestamp = get_timestamp();
```
luckysori, line 987: 🔧  Since it is easy to make a mistake and forget to update the `updated_timestamp` every time we update the payment status, perhaps it would be a good idea to make the fields private and a public method to mutate the state, which automatically updates the `updated_timestamp`.

rust/src/db.rs
```diff
@@ -24,6 +30,15 @@ pub type SqliteConnection = PoolConnection<Sqlite>;
 /// Wallet has to be managed by Rust as generics are not support by frb
 static DB: Storage<Mutex<SqlitePool>> = Storage::new();
 
+fn into_array(vec: Vec<u8>) -> [u8; 32] {
+    let slice = vec.as_slice();
+    let array: [u8; 32] = match slice.try_into() {
+        Ok(arr) => arr,
+        Err(_) => panic!("Expected a Vec of length {} but it was {}", 32, vec.len()),
+    };
+    array
```
luckysori, line 39: 🙃 StackOverflow offers a simpler version: https://stackoverflow.com/a/29570662

```rust
v.try_into()
        .unwrap_or_else(|v| panic!("Expected a Vec of length 32 but it was {}", v.len()))
```

Does that not work?

rust/src/db.rs
```diff
@@ -196,3 +211,196 @@ pub async fn insert_ignore_txid(txid: Txid, maker_amount: i64) -> Result<()> {
...
+        hash,
+        preimage,
+        secret,
+        status,
+        amt_msat,
+        created_timestamp,
+        updated_timestamp,
+    } = payment.clone();
+
+    let query_result = sqlx::query(
```
luckysori, line 229: 🔧 Shouldn't we use the `query!` macro to get some compile-time safety?

rust/src/db.rs
```diff
@@ -196,3 +211,196 @@ pub async fn insert_ignore_txid(txid: Txid, maker_amount: i64) -> Result<()> {
...
+    .bind(created_timestamp as i64)
+    .bind(updated_timestamp as i64)
+    .execute(&mut conn)
+    .await?;
+
+    ensure!(
+        query_result.rows_affected() == 1,
+        "Failed to insert payment"
+    );
+    tracing::info!("Successfully stored new payment");
```
luckysori, line 250: 🙃 We could add the payment hash to these logs.

rust/src/db.rs
```diff
@@ -196,3 +211,196 @@ pub async fn insert_ignore_txid(txid: Txid, maker_amount: i64) -> Result<()> {
...
+    Ok(())
+}
+
+pub async fn update_payment(hash: PaymentHash, status: HTLCStatus) -> Result<()> {
+    let mut connection = acquire().await?;
+    let updated = time::OffsetDateTime::now_utc().unix_timestamp();
```
luckysori, line 256: 🤔 Hmm, no I'm confused. I thought this timestamp was set on the outside?

rust/migrations/20221128012310_add_payments.sql
```diff
@@ -0,0 +1,12 @@
+-- Add lightning payments table
+CREATE TABLE IF NOT EXISTS payments (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    payment_hash TEXT UNIQUE NOT NULL,
+    preimage TEXT,
+    secret TEXT,
+    htlc_status TEXT NOT NULL,
+    amount_msat INTEGER,
```
luckysori, line 9: ❓ How can the amount be null?

da-kami: I guess it depends on what a `payment` is - an `invoice` can be created without an amount (and then the payer chooses the amount). 
For me a payment would identify something that has a concrete amount that is either incoming or outgoing. That said I think this should be `NOT NULL` I guess...?

rust/src/db.rs
```diff
@@ -21,6 +30,15 @@ pub type SqliteConnection = PoolConnection<Sqlite>;
 /// Wallet has to be managed by Rust as generics are not support by frb
 static DB: Storage<Mutex<SqlitePool>> = Storage::new();
 
+fn into_array(vec: Vec<u8>) -> [u8; 32] {
+    let slice = vec.as_slice();
+    let array: [u8; 32] = match slice.try_into() {
+        Ok(arr) => arr,
+        Err(_) => panic!("Expected a Vec of length {} but it was {}", 32, vec.len()),
```
da-kami, line 37: Why `panic` here and not use a `try_from` or similar?

rust/src/db.rs
da-kami: Note: In itchysats we used the `query!` macro for `select` statements, but not for inserts. Not sure what was the reason though.

rust/src/db.rs
da-kami: Yeah, we set this twice and effectively it will be what we set here. I think we have  to agree on what this timestamp represents. Is `updated` when we *received* the payment update, or is it (internally) use to mark when we last updated the row in the db?

da-kami (approved):
Overall approach ack! 
Feel free to merge when the comments are actioned.

rust/src/db.rs
klochowicz: we're not using this new API anywhere yet! so far I just aligned the amount of fields, and the part of actual insertion / update will be changed. :)

rust/migrations/20221128012310_add_payments.sql
klochowicz: unfortunately, it can be null, according to the rust-lightning people ;/ I am keeping it as close as possible to the current model (which actually supports null).

rust/src/db.rs
klochowicz: for me, this is not an error that can happen at runtime - provided that the attached unit test passes, this error condition is weeded out. 
rust-lightning types have to have the slice of that length in them, this API cannot change.

rust/src/db.rs
klochowicz: I just followed the other insert statement, I will see whether both can be `query!` instead. thanks for spotting 🙏

rust/src/db.rs
klochowicz: indeed - great suggestion!

also: I thought anyone would comment on the use of `map()` for value inside the option, I only recently found that out and made things way more readable :D

rust/src/lightning.rs
klochowicz: that is indeed the goal, hence the API proposed had this done internally :)
I will try to hide as much of the public fields away in the next stage.

klochowicz:
the rebase will take quite a while now, due to the large refactor that got in in the meantime 😢

rust/src/db.rs
da-kami: fair enough

da-kami:
> the rebase will take quite a while now, due to the large refactor that got in in the meantime 😢

If you need help let me know.

Note: Reminder that  the `sqlx-data.json` can be fixed by just taking either and then running the `prepare` script again to generate the actual one to be used. No need to try to merge this :)

klochowicz:
> > the rebase will take quite a while now, due to the large refactor that got in in the meantime 😢
> 
> If you need help let me know.
> 
> Note: Reminder that the `sqlx-data.json` can be fixed by just taking either and then running the `prepare` script again to generate the actual one to be used. No need to try to merge this :)

thanks for the offer, I managed to solve this :)

rust/src/db.rs
klochowicz: I couldn't make this work easily, I left it as all the other insert statements as it's not the priority right now.

PR #397, opened 2022-11-28
Let the taker's Bitcoin wallet be drained

luckysori:
If the user specifies a send amount that _equals_ their total funds, we call `drain_to` instead of `add_recipient`. Otherwise it fails because we cannot pay for transaction fees.

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -251,6 +251,8 @@ impl Wallet {
     }
 
     pub fn send_to_address(&self, send_to: Address, amount: u64) -> Result<Txid> {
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Nice. Additionally we could say if the user enters `0` the wallet will be drained.

luckysori:
> Nice. Additionally we could say if the user enters `0` the wallet will be drained.

Similarly, I considered letting the user go over their balance to trigger draining the wallet, but I was worried that this could happen by mistake. I'd be more comfortable if we explicitly let the user click a button to drain the wallet from the send-bitcoin screen.

Issue #389, opened 2022-11-26
Redesign async button


holzeis:
We should have a dedicated (async) button widget to process async task - based on the work done in open channel.

- [ ] Disable button if async action is still processing (preventing premature retries)
- [ ] Show spinning wheel on button if async action is processing
- [ ] Make button larger so its easier to click

PR #398, opened 2022-11-28
Introduce reusable `SubmitButton`

da-kami:
resolves https://github.com/itchysats/10101/issues/389

Button widget that takes a `dynamic` arbitrary function callback and calls it in an async fashion.

On success evaluation of the future we show a success snackbar message. On failure evaluation of the future we show a failure snackbar.

Note:
Dart does not allow arbitrary function arguments as part of type arguments, i.e. we cannot define a function that "takes any amount of dynamic arguments". To allow some level of flexibility we allow functions with 0, 1, 2 and 3 parameters where the parameters are passed as separate optional arguments.

---

Will be done in follow up:

-  Make the button larger (We should handle that through the Theme and not set anything specific in `AsyncButton`).

### DIFF
--- a/lib/cfd_trading/cfd_order_confirmation.dart
+++ b/lib/cfd_trading/cfd_order_confirmation.dart
@@ -126,7 +126,7 @@ class _CfdOrderConfirmationState extends State<CfdOrderConfirmation> {
                 child: Container(
                   alignment: Alignment.bottomRight,
                   child: AsyncButton(
-                    onPressedFunction: () async {
+                    onPressed: () async {
                       await openCfd(order, cfdTradingChangeNotifier);
                     },
                     label: 'Confirm',
--- a/lib/cfd_trading/cfd_order_confirmation.dart
+++ b/lib/cfd_trading/cfd_order_confirmation.dart
@@ -1,6 +1,5 @@
 import 'package:f_logs/model/flog/flog.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:go_router/go_router.dart';
 import 'package:intl/intl.dart';
 import 'package:provider/provider.dart';
@@ -10,6 +9,7 @@ import 'package:ten_ten_one/cfd_trading/validation_error.dart';
 import 'package:ten_ten_one/models/amount.model.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
+import 'package:ten_ten_one/utilities/async_button.dart';
 
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 
@@ -54,7 +54,6 @@ class _CfdOrderConfirmationState extends State<CfdOrderConfirmation> {
     formatter.minimumFractionDigits = 2;
     formatter.maximumFractionDigits = 2;
 
-    final cfdTradingService = context.read<CfdTradingChangeNotifier>();
     final cfdTradingChangeNotifier = context.read<CfdTradingChangeNotifier>();
     Order order = widget.args.order;
     Message? channelError = widget.args.channelError;
@@ -122,40 +121,45 @@ class _CfdOrderConfirmationState extends State<CfdOrderConfirmation> {
                   : AlertMessage(message: channelError),
             ),
             const SizedBox(height: 50),
+            Padding(
+                padding: const EdgeInsets.only(right: 20.0),
+                child: Container(
+                  alignment: Alignment.bottomRight,
+                  child: AsyncButton(
+                    onPressedFunction: () async {
+                      await openCfd(order, cfdTradingChangeNotifier);
+                    },
+                    label: 'Confirm',
+                    isButtonDisabled: channelError == null,
+                  ),
+                )),
           ]),
         ),
       ),
-      floatingActionButton: ElevatedButton(
-          onPressed: channelError == null
-              ? () async {
-                  try {
-                    await api.openCfd(order: order);
-                    FLog.info(text: 'OpenCfd returned successfully');
-
-                    // switch index to cfd overview tab
-                    cfdTradingChangeNotifier.selectedIndex = 1;
-
-                    // refreshing cfd list after cfd has been opened
-                    // will also implicitly propagate the index change
-                    await cfdTradingService.refreshCfdList();
-
-                    // navigate back to the trading route where the index has already been propagated
-                    context.go(CfdTrading.route);
-                  } on FfiException catch (error) {
-                    FLog.error(text: 'Failed to open CFD: ' + error.message, exception: error);
-
-                    ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
-                      backgroundColor: Colors.red,
-                      content: Text("Failed to open cfd"),
-                    ));
-
-                    return;
-                  }
-
-                  context.go(CfdTrading.route);
-                }
-              : null,
-          child: const Text('Confirm')),
     );
+  }
+
+  Future<void> openCfd(Order order, CfdTradingChangeNotifier cfdTradingChangeNotifier) async {
+    FLog.info(text: "Opening CFD with order " + order.toString());
+    await api.openCfd(order: order).then((value) async {
+      ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
+        content: Text("CFD opened"),
+      ));
+
+      // switch index to cfd overview tab
+      cfdTradingChangeNotifier.selectedIndex = 1;
+
+      // refreshing cfd list after cfd has been opened
+      // will also implicitly propagate the index change
+      await cfdTradingChangeNotifier.refreshCfdList();
+
+      context.go(CfdTrading.route);
+    }).catchError((error) {
+      FLog.error(text: "Failed to open CFD.", exception: error);
+      ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+        backgroundColor: Colors.red,
+        content: Text("Failed to open CFD. Error: " + error.toString()),
+      ));
+    });
   }
 }
--- a/lib/cfd_trading/cfd_order_confirmation.dart
+++ b/lib/cfd_trading/cfd_order_confirmation.dart
@@ -9,7 +9,7 @@ import 'package:ten_ten_one/cfd_trading/validation_error.dart';
 import 'package:ten_ten_one/models/amount.model.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
-import 'package:ten_ten_one/utilities/async_button.dart';
+import 'package:ten_ten_one/utilities/submit_button.dart';
 
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 
@@ -125,7 +125,7 @@ class _CfdOrderConfirmationState extends State<CfdOrderConfirmation> {
                 padding: const EdgeInsets.only(right: 20.0),
                 child: Container(
                   alignment: Alignment.bottomRight,
-                  child: AsyncButton(
+                  child: SubmitButton(
                     onPressed: () async {
                       await openCfd(order, cfdTradingChangeNotifier);
                     },
--- a/lib/cfd_trading/cfd_order_detail.dart
+++ b/lib/cfd_trading/cfd_order_detail.dart
@@ -9,7 +9,7 @@ import 'package:ten_ten_one/cfd_trading/cfd_trading_change_notifier.dart';
 import 'package:ten_ten_one/cfd_trading/validation_error.dart';
 import 'package:ten_ten_one/models/amount.model.dart';
 import 'package:ten_ten_one/models/order.dart';
-import 'package:ten_ten_one/utilities/async_button.dart';
+import 'package:ten_ten_one/utilities/submit_button.dart';
 import 'package:ten_ten_one/utilities/tto_table.dart';
 import 'package:go_router/go_router.dart';
 
@@ -173,7 +173,7 @@ class _CfdOrderDetailState extends State<CfdOrderDetail> {
                             ),
                             Visibility(
                               visible: confirm,
-                              child: AsyncButton(
+                              child: SubmitButton(
                                   onPressed: () async {
                                     await settleCfd(cfd, offer, cfdTradingChangeNotifier);
                                   },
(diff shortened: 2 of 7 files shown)


lib/wallet/open_channel.dart
```diff
@@ -137,48 +136,20 @@ class _OpenChannelState extends State<OpenChannel> {
...
-                            height: 18,
-                            padding: const EdgeInsets.all(2.0),
-                            child: const CircularProgressIndicator(
-                              color: Colors.white,
-                              strokeWidth: 3,
-                            ),
-                          ),
-                    onPressed: !validForm || submitting ? null : openChannel),
+                child: AsyncButton(
+                  onPressedFunction: context.read<ChannelChangeNotifier>().open,
```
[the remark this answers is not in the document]
holzeis: also I would call it onPressed instead of onPressedFunction then 😉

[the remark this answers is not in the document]
da-kami: I force pushed, please have a look again @holzeis :)

lib/utilities/async_button.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:flutter/material.dart';
+
+class AsyncButton extends StatefulWidget {
+  final void Function() onPressedFunction;
+  final String label;
+
+  const AsyncButton({required this.onPressedFunction, required this.label, Key? key})
```
holzeis: 🔧 We need to be able to disable the button from the outside, e.g. if the form is invalid. This was already done on the open channel button, e.g. if the channel amount was not set.

lib/utilities/async_button.dart
```diff
@@ -0,0 +1,46 @@
...
+                ),
+              ),
+        onPressed: submitting
+            ? null
+            : () {
+                setState(() {
+                  submitting = true;
+                });
+
+                widget.onPressedFunction();
```
holzeis: 🤔 Isn't it kind of risky keeping the error handling outside, as we would not be able to set the submitting state to false again if the calling widget does not properly handle the error.

I suggest we wrap that into a try catch and re-throw the exception when the action fails. This way we can add our error handling.

lib/wallet/channel_change_notifier.dart
```diff
@@ -7,7 +7,7 @@ class ChannelChangeNotifier extends ChangeNotifier {
   bool _init = false;
   bool isInitialising() => _init;
 
-  Future<void> open(int amount) async {
+  Future<void> open(dynamic amount) async {
```
holzeis, line 10: ❓ is this dynamic still needed here?

lib/cfd_trading/cfd_order_confirmation.dart
```diff
@@ -13,6 +12,8 @@ import 'package:ten_ten_one/utilities/tto_table.dart';
 
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
 
+import '../utilities/async_button.dart';
```
holzeis, line 15: 🔧 Please use package imports, that will allow us to move the button to another directory without having to update this file.

lib/cfd_trading/cfd_order_confirmation.dart
```diff
@@ -122,40 +123,42 @@ class _CfdOrderConfirmationState extends State<CfdOrderConfirmation> {
                   : AlertMessage(message: channelError),
             ),
             const SizedBox(height: 50),
+            Padding(
+                padding: const EdgeInsets.only(right: 20.0),
+                child: Container(
+                  alignment: Alignment.bottomRight,
+                  child: AsyncButton(
+                    onPressedFunction: openCfd,
```
holzeis: 🔧 The button should be disabled if there is a channel error.

lib/wallet/open_channel.dart
```diff
@@ -137,40 +138,24 @@ class _OpenChannelState extends State<OpenChannel> {
...
-                            width: 18,
-                            height: 18,
-                            padding: const EdgeInsets.all(2.0),
-                            child: const CircularProgressIndicator(
-                              color: Colors.white,
-                              strokeWidth: 3,
-                            ),
-                          ),
-                    onPressed: !validForm || submitting ? null : openChannel),
+                child: AsyncButton(
```
holzeis, line 141: 🔧 The button should be disabled if the form is invalid.

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -176,32 +176,7 @@ class _CfdOrderDetailState extends State<CfdOrderDetail> {
...
-                                          exception: error);
-
-                                      ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
-                                        backgroundColor: Colors.red,
-                                        content: Text("Failed to settle cfd"),
-                                      ));
-                                    }
-                                  },
-                                  child: const Text('Confirm')),
+                              child: AsyncButton(onPressedFunction: settleCfd, label: 'Confirm'),
```
holzeis: nit: we don't have to put the offer as late state variable.

```
AsyncButton(onPressedFunction: () { settleCfd(offer) }, label: 'Confirm')
```

lib/wallet/send_on_chain.dart
```diff
@@ -74,31 +74,28 @@ class _SendOnChainState extends State<SendOnChain> {
...
                         ),
                       ],
                     ),
                   ),
                 ]))));
   }
+
+  void sendOnChain() {
+    FLog.info(text: "Sending $amount to $address");
+    api.sendToAddress(address: address, amount: amount).then((value) async {
```
holzeis, line 87: ❓ Shouldn't we add an await here somewhere?

lib/wallet/send.dart
```diff
@@ -53,30 +53,28 @@ class _SendState extends State<Send> {
...
                         ),
                       ],
                     ),
                   ),
                 ]))));
   }
+
+  void send() {
+    FLog.info(text: "Paying invoice $invoice");
+    api.sendLightningPayment(invoice: invoice).then((value) async {
```
holzeis, line 66: ❓ Shouldn't we add an await here somewhere?

da-kami: I am happy to stick to the `await` and `try ... catch` syntax - I don't know why we introduced different syntax using the `.then` combinator.

either way we should stick to one pattern I think. At the moment it is confusing, because in some places we have `try ...  await ...  catch` and in others we have `.then ... .error`.

lib/wallet/send_on_chain.dart
da-kami: With this syntax it is not needed - but I'm happy to stick to `try ... await ... catch`. I thought that you introduced the `.then` syntax and gave it a try 😅

lib/cfd_trading/cfd_order_detail.dart
da-kami: Fair enough, I'll change that :)

lib/wallet/send.dart
luckysori: According to @holzeis, if you use `.then` instead of `await`, you don't wait for the future to resolve.

lib/cfd_trading/cfd_order_detail.dart
```diff
@@ -50,10 +52,8 @@ class _CfdOrderDetailState extends State<CfdOrderDetail> {
     formatter.minimumFractionDigits = 2;
     formatter.maximumFractionDigits = 2;
 
-    final cfdTradingService = context.read<CfdTradingChangeNotifier>();
-
     final cfdOffersChangeNotifier = context.watch<CfdOfferChangeNotifier>();
```
luckysori, line 55: 🙃 I guess we should align the variable name with the type.


```suggestion
    final cfdOfferChangeNotifier = context.watch<CfdOfferChangeNotifier>();
```

lib/utilities/async_button.dart
```diff
@@ -0,0 +1,45 @@
...
+
+class _AsyncButtonState extends State<AsyncButton> {
+  bool submitting = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return ElevatedButton.icon(
+        label: Text(widget.label),
+        icon: !submitting
+            ? Container()
```
luckysori, line 24: 🤔 For some reason using an empty `Container` didn't work for me in the context of [closing the channel](https://github.com/itchysats/10101/pull/394#discussion_r1033462062). We might need to use a `SizedBox.shrink()` instead?

luckysori (approved):
LGTM!

lib/cfd_trading/cfd_order_detail.dart
da-kami: Note: I also adapted this to passing parameters rather than adding fields to the state class for `cfd_order_confirmation`

lib/wallet/send.dart
da-kami: Hm... that's not how I would understand [the docs](https://api.flutter.dev/flutter/dart-async/Future/then.html). My understanding is that `then` is only called when the future resolves, which implies for me that it is awaited.

I find the `try ... catch` syntax more readable anyway, so I'll change this.

luckysori: Okay, this is what Richie was referring to: https://stackoverflow.com/questions/37551843/invoking-async-function-without-await-in-dart-like-starting-a-thread.

As I understand it, if you omit the `await` the future runs, but you continue and don't care about the future.

That is why you'd want to attach combinators in that case, to handle what happens without waiting for it to happen!

---

This is obviously different to how futures work in Rust, where you have to `await` if you want the future to be polled.

lib/cfd_trading/cfd_order_detail.dart
da-kami: happy to do that in a follow-up, this leaks through many files and is unrelated to this PR

lib/wallet/send.dart
luckysori: Interesting caveat from that thread:

> This isn't like a thread though. As mentioned processing is enqueued to the event queue which means it won't be executed until the current task and all previously enqueued tasks are completed.
>
> If you want real parallel execution you need to utilize isolates.

lib/utilities/async_button.dart
da-kami: I had this problem when using the `ElevatedButton` (i.e. `AsyncButton`) as FAB, but I think we actually should not use it as FAB in our context, because what we have is more of a form submission rather than triggering an action.
I ran into this problem here as well: https://github.com/itchysats/10101/pull/398/commits/7a2b10645f283f5276c458abb9af0f67168cbbeb and fixed it by adding the Button to the body instead of using a FAB.

I am not sure this is 100% "correct", but I think the FAB makes sense when we have (one or multiple) action that apply to the screen. Once we are in a sub-screen where the user submits a form, I think it  makes more sense to integrate the button in the body as form submission.

luckysori: That makes sense, thanks for the explanation!

lib/wallet/send.dart
da-kami: Ah, OK! I think I was confused by the syntax, but I can add the `await` regardless of the syntax to await the *returned* `Future`! Thanks for the link, that helped. 

We can actually keep the syntax, but add the `await`, from what I understand from the link you sent it will indeed change the behavior. 
I wonder what is the most idiomatic way to await and lifecycle management of dart futures in general. I could also return a future from the helper functions I created and then `await` when calling them on the `onPressed` function of the button - would we expect an `await` there?

da-kami:
Everything should be fixed and aligend now!

Took me a bit to get the dart `Future`s right - the button takes a function that expects a `Future` to be returned (`Future<void` because we don't expect a value to be returned). The button awaits the future - and `async` is properly depiced in the caller as well, pulled through all call levels.

lib/utilities/submit_button.dart
```diff
@@ -0,0 +1,52 @@
...
+            : () async {
+                setState(() {
+                  submitting = true;
+                });
+
+                try {
+                  await widget.onPressed();
+                } on Exception {
+                  rethrow;
+                } finally {
```
luckysori, line 45: ❓Can you explain this bit?

lib/wallet/open_channel.dart
```diff
@@ -137,40 +138,25 @@ class _OpenChannelState extends State<OpenChannel> {
...
   }
 
-  void openChannel() {
-    setState(() {
-      submitting = true;
-    });
+  Future<void> openChannel() async {
     FLog.info(text: "Opening Channel with capacity " + takerChannelAmount.toString());
-    context.read<ChannelChangeNotifier>().open(takerChannelAmount).then((value) {
+    await context.read<ChannelChangeNotifier>().open(takerChannelAmount).then((value) {
```
luckysori, line 154: ❓ Do we want to change this behaviour?

I know we talked about generally always awaiting, but I think this was left without an `await` on purpose by @holzeis.

lib/wallet/send.dart
```diff
@@ -53,30 +53,28 @@ class _SendState extends State<Send> {
...
                         ),
                       ],
                     ),
                   ),
                 ]))));
   }
+
+  Future<void> send() async {
+    FLog.info(text: "Paying invoice $invoice");
+    await api.sendLightningPayment(invoice: invoice).then((value) async {
```
luckysori, line 66: ❓ Does the `then` callback need to be async?

luckysori (approved):
LGTM, except for maybe the change of behaviour in open-channel.

lib/utilities/submit_button.dart
da-kami: `finally` is executed either way, i.e. it is executed after `try` finishes but it is also executed after `catch` in case of an error.

This is a common exception handling pattern in Java as well - it works pretty much the same in `dart` (there can be tiny differences in behavior when returning from e.g. a finally block, but one should just not do that...).

If it is confusing I can also just do a `try ... catch` and repeat the code of setting `submitting` to false :)

lib/wallet/open_channel.dart
da-kami: If that was *on purpose* I think it really deserves a comment :)
Happy to keep the previous behavior and I'm adding a comment. I suppose we don't `await` here because the `Open Channel` card shows the status.

lib/wallet/send.dart
da-kami: Since we are not doing `async` operations in this `then` block we don't need it, nice spotting :)
I removed it for `send_on_chain` as well :)

lib/utilities/submit_button.dart
luckysori: I highlighted the wrong bit haha. Sorry!

My question was about why we are `rethrow`ing and what we expect to happen then.

lib/wallet/open_channel.dart
luckysori: Agree on adding a comment!

luckysori (approved):
Very approved.

lib/wallet/open_channel.dart
holzeis: @da-kami @luckysori I guess the await is not needed anymore here, since the SubmitButton is awaiting on the complete open_channel function, so this await has simply moved 😉

PR #399, opened 2022-11-28
fix: make ln payment work

### LINKED ISSUE #313: Pending after receiving over lightning does not go away

bonomat:
We don't know exaclty why it was not working before, but this works :)

resolves #313


holzeis (approved):
I've already noticed, that we have a very limited number of threads on the phone. Could it be that we ended up with too many threads before (since we did not await)?

PR #401, opened 2022-11-28
fix: use action to add issue to board

bonomat:


bonomat:
I'm going ahead and merge this rightaway

PR #434, opened 2022-11-29
Sync the maker with the blockchain periodically

luckysori:
We split syncing from getting the balance in patch f7e300dbb31fca4f1aadc1504cc711365220909e. This was okay for the taker, who still syncs periodically.

On the other hand, the maker was only calling `get_balance` periodically. Therefore, the UTXO set was no longer being updated.

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -47,6 +47,11 @@ async fn main() -> Result<()> {
 
         loop {
             let started = Instant::now();
+
+            if let Err(e) = wallet::sync() {
+                tracing::error!("Wallet sync failed: {e:#}");
+            }
+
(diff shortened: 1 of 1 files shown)


da-kami (approved):

holzeis (approved):

bonomat:
> We split syncing from getting the balance in patch [f7e300d](https://github.com/itchysats/10101/commit/f7e300dbb31fca4f1aadc1504cc711365220909e). This was okay for the taker, who still syncs periodically.
> 
> On the other hand, the maker was only calling `get_balance` periodically. Therefore, the UTXO set was no longer being updated.

woops, sorry 🙈

PR #439, opened 2022-11-29
feat: Add invoice details

### LINKED ISSUE #408: Show invoice details when paying an invoice
Currently the user can only add the invoice as text, without any information about the contents of the invoice. At a very minimum the user should be able to 

- [ ] review the amount requested in the invoice 
- [ ] and counter party

holzeis:
invoice-details

resolves #408

### DIFF
--- a/lib/wallet/send.dart
+++ b/lib/wallet/send.dart
@@ -1,8 +1,12 @@
 import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
+import 'package:intl/intl.dart';
+import 'package:ten_ten_one/bridge_generated/bridge_definitions.dart';
+import 'package:ten_ten_one/cfd_trading/validation_error.dart';
 import 'package:ten_ten_one/ffi.io.dart';
 import 'package:ten_ten_one/utilities/submit_button.dart';
+import 'package:ten_ten_one/utilities/tto_table.dart';
 
 class Send extends StatefulWidget {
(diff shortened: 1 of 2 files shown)


bonomat (approved):
Very nice :)

lib/wallet/send.dart
```diff
@@ -24,6 +29,83 @@ class _SendState extends State<Send> {
...
+          try {
+            invoice = await api.decodeInvoice(invoice: encodedInvoice);
+            setState(() {
+              invoice = invoice;
+            });
+          } catch (error) {
+            FLog.error(text: "Failed to decode invoice $encodedInvoice.", exception: error);
+            ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+              backgroundColor: Colors.red,
+              content: Text("Failed to decode invoice. Error: " + error.toString()),
```
luckysori, line 53: 🔧 I think the user likely cannot make sense of the extra error context or do anything with that extra information. I would omit it and just say that they provided an invalid invoice.

luckysori (approved):
I really like how it's responsive to what the user pastes in!

lib/wallet/send.dart
klochowicz: different perspective: IMHO at this stage of our product, any user feedback is helpful - if the user reports odd error message, then it's still good for us. And we also get more data doing internal testing, which is annoying, especially on mobile...

klochowicz:
I'll go ahead with it, as I'm adding the QR code scanning feature on top of this.

PR #442, opened 2022-11-29
Align safe area

da-kami:
Notes on setting `SystemUiOverlayStyle`:

I wanted to achieve that for the main screens we would use `SystemUiOverlayStyle.dark` and for the child screens (with blue app-bar on top) we would use `SystemUiOverlayStyle.light`, but unfortunately this is super tricky. I cold not get it work properly.
Unfortunately the behavior of setting the `SystemUiOverlayStyle` is not consistent with the docs - at least on the iOS emulator (I have to assume it will behave the same on iOS...).

I tried:

- Set `SystemUiOverlayStyle.dark` in `app_bar_with_balance.dart`'s `AppBar` - so that this one is always dark. But unfortunately whenever I navigate to a child screen (e.g. Receive, CFD Order, ...) this resets `SystemUiOverlayStyle.light`. When navigating back the status bar is white and not visible. I assumed this is because of the default value.
- I tried numerous approaches to get the balance / main screens to use `SystemUiOverlayStyle.dark`, but I always ran into the same problem. (I wrapped it with a Theme that sets it, I tried to wrap a Theme around the Scaffold rather than the balance, ...)
- I came to the conclusion that I have to change the default to `SystemUiOverlayStyle.dark`. This works, but only as long as I don't set it to `SystemUiOverlayStyle.light` in one of the child screens. If I do that it somehow overwrites the default (! - I think that's a bug...) and when navigating back the status bar is invisible because white again. 
- I tried to set the `SystemUiOverlayStyle` where we define the routes (would be a bit ugly, but at least in one place...), but that is tricky as well, because we would have to use `ShellRoutes` to achieve that and that blows up complexity...

I have a hunch that this is related to the theme set by the Swatch, but I already wasted enough time so...
I gave up and just accepted the solution where we set it to `dark` globally, but this does not look especially good on child screens 😅

### DIFF
--- a/lib/cfd_trading/cfd_trading.dart
+++ b/lib/cfd_trading/cfd_trading.dart
@@ -32,40 +32,39 @@ class _CfdTradingState extends State<CfdTrading> {
 
     const balanceSelector = BalanceSelector.lightning;
 
-    return SafeArea(
-      child: Scaffold(
-          appBar: PreferredSize(
-              child: const AppBarWithBalance(balanceSelector: balanceSelector),
-              preferredSize: Size.fromHeight(balanceSelector.preferredHeight)),
-          drawer: const Menu(),
-          body: _pages.elementAt(cfdTradingService.selectedIndex),
-          bottomNavigationBar: BottomNavigationBar(
-            items: const <BottomNavigationBarItem>[
-              BottomNavigationBarItem(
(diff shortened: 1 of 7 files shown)


bonomat (approved):
I guess this works 🤷‍♂️ :)

Issue #352, opened 2022-11-24
Hide deposit bitcoin card after we see funding transaction in mempool


da-kami:
We see the `Deposit Bitcoin` card until the transaction is confirmed which can be quite long. We could hide it.

This is related to https://github.com/itchysats/10101/issues/337 because it's a similar problem
This is related to #347 because if we handle the on-chain balance more fine granular

Issue #347, opened 2022-11-24
taker balance is 0 before `open_channel` succeds


klochowicz:
Even if funded previously, the taker's balance is 0 until the other transaction gets confirmed:

![image](https://user-images.githubusercontent.com/8319440/203690447-0baf8dd8-b331-449b-9a2e-b6e64eac64d0.png)

da-kami:
This is because we use `confirmed` balance to display the on-chain balance, and with only one unspent output (that is then used for opening the channel) you don't have any confirmed BTC until the transaction is confirmed. 

The *proper* solution for this is to have an indicator on each balance that shows if the balance is confirmed or not, and display unconfirmed balance in between. This, however, will require some design, because it can become cluttered with yet another dimension of indication on the on-chain balance. We could give it a try though.

PR #446, opened 2022-11-30
Take into account pending on-chain balance

luckysori:
Fixes https://github.com/itchysats/10101/issues/352.

Depending on the situation we make use of different values:

- The balance widget now displays the total balance by default, but on hover one can see a tooltip with the breakdown between confirmed and pending.
- We use the confirmed on-chain balance to determine how much bitcoin they user can use to either open a channel or send on-chain.
- We use the total on-chain balance to determine if the initial deposit-bitcoin card should be rendered or not.

image

---

The first commit (https://github.com/itchysats/10101/commit/f56ab4895d9f2fbb20f4e36f00ccdc719cf970e4) was useful for testing so I decided to keep it around.

### DIFF
--- a/lib/balance.dart
+++ b/lib/balance.dart
@@ -15,14 +15,29 @@ class Balance extends StatelessWidget {
   Widget build(BuildContext context) {
     return Consumer2<LightningBalance, BitcoinBalance>(
       builder: (context, lightningBalance, bitcoinBalance, child) {
-        var bitcoinBalanceDisplay = bitcoinBalance.amount.display(currency: Currency.sat);
+        var bitcoinBalanceTotalDisplay = bitcoinBalance.total().display(currency: Currency.sat);
+        var bitcoinBalanceConfirmedDisplay =
+            bitcoinBalance.confirmed.display(currency: Currency.sat);
+        var bitcoinBalancePendingDisplay = bitcoinBalance.pending().display(currency: Currency.sat);
+
         var lightningBalanceDisplay = lightningBalance.amount.display(currency: Currency.sat);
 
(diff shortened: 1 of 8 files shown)


da-kami (commented):
Does this fix https://github.com/itchysats/10101/issues/347 ?
Could you add a screenshot? It's a bit hard to imaging how this looks and feels. You mention that "on hover" it displays pending balance - where is the code for this and what would "hover" mean on the mobile phone? :)

luckysori:
> Does this fix #347 ? 

> Could you add a screenshot? 

Woops, forgot to insert it before. It should be there now.

> You mention that "on hover" it displays pending balance - where is the code for this and what would "hover" mean on the mobile phone? :)

Great question. According to the documentation of `Tooltip`:

> Wrap the button in a [Tooltip](https://api.flutter.dev/flutter/material/Tooltip-class.html) widget and provide a message which will be shown when the widget is long pressed.

bonomat (approved):
That's a nicer solution than mine in : 85f338400682482c30643812907cac18c182b3c5 :)

bonomat:
> That's a nicer solution than mine in : [85f3384](https://github.com/itchysats/10101/commit/85f338400682482c30643812907cac18c182b3c5) :)

I've dropped mine in favor of this.

klochowicz:
@luckysori FYI this might fail on bors due to  lack of `make fmt` before submitting

luckysori:
> @luckysori FYI this might fail on bors due to lack of `make fmt` before submitting

Good call. The rebase conflicts broke the formatting :(

luckysori:
bors retry

PR #453, opened 2022-11-30
Feat: maker faucet

### LINKED ISSUE #406: Fund the taker for demo purposes
In order to mitigate the risk for a user to close the app again, because she doesn't have any testnet money - an option should be added allowing the taker to requests funds from the maker for testing.

**Acceptance Criterias**

- [ ] Funding works only one time!
- [ ] Funds are sent to the wallet address managed by the wallet. about 10.000 sats should be enough
- [ ] Additional tip added, explaining how to get to more testnet money.
- [ ] Add disclaimer that this is only for testing purposes and will be removed in a productive solution.

bonomat:
resolves #406 

I followed the design of https://bitcoin.design/guide/daily-spending-wallet/first-use/#initial-deposit a bit.

Doesn't look as sexy and I probably broke all laws of flutter design, but I like it :D

Video is a bit longer because it didn't sync :) 

https://user-images.githubusercontent.com/224613/204707420-e8d01ee5-9c4c-4850-b1e6-18a441699b54.mov


rust/src/faucet.rs
```diff
@@ -0,0 +1,35 @@
...
+    let result = client
+        .get(maker_endpoint() + "/api/faucet/" + address.as_str())
+        .send()
+        .await;
+    let response = match result {
+        Ok(res) => res,
+        Err(err) => {
+            tracing::error!("Could not call faucet {err:?}");
+            return Ok(None);
+        }
```
luckysori, line 20: ❓ Why not return an error here? What's the `Ok(None)` case for?

lib/wallet/fund_wallet_on_chain.dart
```diff
@@ -0,0 +1,129 @@
...
+                  const Divider(),
+                  InkWell(
+                      onTap: () {
+                        Navigator.push(
+                          context,
+                          MaterialPageRoute(builder: (context) => const ReceiveOnChain()),
+                        );
+                      },
+                      child: const ListTile(
+                        title: Text("Receive bitcoin"),
```
luckysori, line 113: I think it's better to use the word "deposit" in this special context. 

```suggestion
                        title: Text("Deposit bitcoin"),
```

lib/wallet/fund_wallet_on_chain.dart
```diff
@@ -0,0 +1,129 @@
...
+                  const Divider(),
+                  InkWell(
+                      onTap: () {
+                        Navigator.push(
+                          context,
+                          MaterialPageRoute(builder: (context) => const DrainFaucet()),
+                        );
+                      },
+                      child: const ListTile(
+                        title: Text("Drain faucet"),
```
luckysori, line 101: 🙃 This might just be me, but drain makes me think that you are going to take all of the coins haha. My suggestion is just an alternative that I don't love either.

```suggestion
                        title: Text("Use faucet"),
```

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                    children: [
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
+                            message: Message(
+                                title: "Use our faucet",
+                                details:
+                                    "By clicking the button below, we will send you 10,000 sats. \n\n"
+                                    "To not get bankrupt, we would appreciate if you do this only once and return us the funds when you are done testing. \n\n"
```
luckysori, line 64: 🔧 

```suggestion
                                    "To not go bankrupt, we would appreciate if you do this only once and return us the funds when you are done testing. \n\n"
```

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                  child: Column(
+                    children: [
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
+                            message: Message(
+                                title: "Use our faucet",
+                                details:
+                                    "By clicking the button below, we will send you 10,000 sats. \n\n"
```
luckysori, line 63: 🙃 Not sure why, but it sounds better 🤷 

```suggestion
                                    "By clicking the button below, you will receive 10,000 sats from us. \n\n"
```

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+          child: address.isEmpty
+              ? const CircularProgressIndicator()
+              : Container(
+                  padding: const EdgeInsets.all(20.0),
+                  child: Column(
+                    children: [
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
```
luckysori, line 59: 🙃 `AlertMessage` seems like the wrong widget, but perhaps others disagree. I would expect this to be text without a special container.

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                            style: TextStyle(color: Colors.black),
+                          ),
+                          TextSpan(
+                            text: 'https://coinfaucet.eu/en/btc-testnet/',
+                            style: const TextStyle(color: Colors.blue),
+                            recognizer: TapGestureRecognizer()
+                              ..onTap = () {
+                                launchUrl(_url);
+                              },
+                          ),
```
luckysori, line 83: 🙃 And then this could be a little informational alert below?

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
+                            message: Message(
+                                title: "Use our faucet",
+                                details:
+                                    "By clicking the button below, we will send you 10,000 sats. \n\n"
+                                    "To not get bankrupt, we would appreciate if you do this only once and return us the funds when you are done testing. \n\n"
+                                    "You can always go to the faucet yourself see below.",
```
luckysori, line 65: 🔧 Maybe better to just combine this with the link below.

```suggestion
                                    "You can always go to the faucet yourself, see below.",
```

luckysori (approved):
Really cool! I like how it was integrated with the deposit-bitcoin card.

lib/wallet/drain_faucet.dart
bonomat: I like the info thingy next to it though :)

lib/wallet/drain_faucet.dart
bonomat: Then we have two on this board :)

rust/src/faucet.rs
bonomat: Fair point, fixed it.

bonomat:
I'm a bit afraid of merge conflicts with this change, so I go ahead and merge this. Feel free to review, I'll see to create a follow-up PR. 

bors r+

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                    children: [
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
+                            message: Message(
+                                title: "Use our faucet",
+                                details:
+                                    "By clicking the button below, you will receive 10,000 sats from us. \n\n"
+                                    "To not go bankrupt, we would appreciate if you do this only once and return us the funds when you are done testing. \n\n"
```
da-kami, line 64:
```suggestion
                                    "To not go bankrupt, we would appreciate if you do this only once and return the funds to us when you are done testing. \n\n"
```

lib/wallet/drain_faucet.dart
```diff
@@ -0,0 +1,111 @@
...
+                  padding: const EdgeInsets.all(20.0),
+                  child: Column(
+                    children: [
+                      const SizedBox(height: 15),
+                      Padding(
+                        padding: const EdgeInsets.only(bottom: 20.0),
+                        child: AlertMessage(
+                            message: Message(
+                                title: "Use our faucet",
+                                details:
```
da-kami, line 62: 🛠️ We have to add the `address` of our maker here so they can return the funds! 

We could actually integrate that in `send-on-chain`, i.e. make it easy for the user to return the funds to us!

lib/wallet/fund_wallet_on_chain.dart
da-kami: If we change `drain` to something else it would be great to update it throughout - we use it in the route as well (...)

lib/wallet/fund_wallet_on_chain.dart
```diff
@@ -0,0 +1,129 @@
...
+          child: Container(
+        padding: const EdgeInsets.all(20.0),
+        child: Column(
+          children: [
+            SizedBox(
+                child: Container(
+              margin: const EdgeInsets.all(20),
+              padding: const EdgeInsets.all(10),
+              decoration: BoxDecoration(
+                color: Colors.orange,
```
da-kami, line 56: 🙃 I think that's the only place in the whole app where we use `Orange` now. I would probably stick to blue 🤷‍♂️

da-kami (approved):
I like the idea! 

From a flow point of view: I think the choice of `Receive` vs. `Use our faucet` is not in line with the information provided. The URL link to the faucet is useful in the `Receive` screen, because that's where one needs the faucet (to paste the address in). It does not help much in the "Use our faucet" screen. But it might be hard to change that, so feel fee to leave as is.

da-kami:
@bonomat even though already borsed it would be great if you could have a look at the comments :)

lib/wallet/drain_faucet.dart
bonomat: Yea, I left a todo because I did not know where to add :D Thought about adding it to the sending fob.

bonomat:
> From a flow point of view: I think the choice of Receive vs. Use our faucet is not in line with the information provided. 

> The URL link to the faucet is useful in the Receive screen, because that's where one needs the faucet (to paste the address in). It does not help much in the "Use our faucet" screen. But it might be hard to change that, so feel fee to leave as is.

I did not want to break the design there. It's not optimal but the flow worked for me because you can go back from the `Receive` screen and click the faucet. 

bors retry

Issue #405, opened 2022-11-28
[CFD Offer] Change quantity to input field and increase leverage options


holzeis:
- [ ] The quantity should be an input field allowing any input between 1 and 1000.
- [ ] Leverage should be selectable up to x10

holzeis:
see also https://github.com/itchysats/10101/issues/372

PR #454, opened 2022-11-30
Quantity text field, validation and more leverage choices

da-kami:
resolves https://github.com/itchysats/10101/issues/405

Took me a while to get this together. The usual "tiny details make the difference" UI challenges :)

I first experimented with changing the screen and moving the `FormTextField` in a separate row. But we run into space problems that are difficult to solve then. 
We could change the *overall* table font size to gain a bit more space, then it might be possible. But I felt it was not worth investing more time into this (it leaks through multiple screens then), so I opted for keeping the three fields in 

I also looked into adding a modal drawer that opens from the bottom where the user is guided to enter the values. But that does not make it much better, because it's hard to see the effect of entering the values then.

Note: I opted for a min quantity of `10` rather than `1` because I think it looks better. The ticket specified `1`, if we insist on that I can change it :)

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/services.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_order_confirmation.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
@@ -26,13 +27,18 @@ class CfdOffer extends StatefulWidget {
 class _CfdOfferState extends State<CfdOffer> {
   late Order order;
 
+  static const minQuantity = 1;
+  static const maxQuantity = 1000;
+
+  final _formKey = GlobalKey<FormState>();
+
   @override
   void initState() {
     super.initState();
 
     order = Order(
         openPrice: 0,
-        quantity: 100,
+        quantity: 10,
         leverage: 2,
         contractSymbol: ContractSymbol.BtcUsd,
         position: Position.Long,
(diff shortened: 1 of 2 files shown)


da-kami:
Only when watching the gif one more time I noticed that `margin` is not properly reset to `0` when the quantity field is empty. Fixed with https://github.com/itchysats/10101/pull/454/commits/d04645d4f448574d06c7539d788818eeb70564cc

lib/cfd_trading/cfd_offer.dart
```diff
@@ -86,96 +89,150 @@ class _CfdOfferState extends State<CfdOffer> {
...
+                children: [
+                  Expanded(
+                    child: TextFormField(
+                      decoration: InputDecoration(
+                        contentPadding: const EdgeInsets.only(bottom: 0.0),
+                        alignLabelWithHint: true,
+                        enabledBorder: UnderlineInputBorder(
+                            borderSide: BorderSide(
+                                color: Theme.of(context).colorScheme.primary, width: 2.0)),
+                        labelText: "* Quantity (10..1000)",
```
luckysori, line 146: 🙃 I would omit the range. Maybe it can be displayed below the text field when the user has it selected?

luckysori (approved):
I approve because I like the gif.

da-kami:
> I'm assuming that the overflow in the gif has nothing to do with this PR.

Indeed it has not, it's the price that sometimes (depending on the price numbers...) overflows on iPhone SE :)

lib/cfd_trading/cfd_offer.dart
da-kami: > Maybe it can be displayed below the text field when the user has it selected?

That's tricky, because the state of the `FormTextField` would become quite complex. It then has to decide if the error is to replace the (second) hint. It certainly does not do that out of the box :)

luckysori: Haha I thought so.

Maybe the range is fine, let's see what the others think.

holzeis (approved):
LGTM, after the tournament we should really start to think how we can better separate the presentation from the logic.

lib/cfd_trading/cfd_offer.dart
bonomat: The notion `(10..1000)` might not be understandable for normal users, what do you think of changing it to `Min:10, Max: 1000`?

Second thought: do we need this label? If we have the one below, that should be good enough I guess?

bonomat (approved):
Looks good on the gif.

lib/cfd_trading/cfd_offer.dart
luckysori: I think you are right with your second thought. It's probably enough if the user gets told the min and max when they input something invalid.

da-kami:
bors r+

Note: I changed the min amount to `1` contract after all, because otherwise test users might run into the problem that they don't have enough sats to open a position.

PR #455, opened 2022-11-30
fix:disable dependabot

bonomat:
#449 was meant to disable dependabot. While doing so it caused an error because the file is now invalid. According to the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit), to disable dependabot temporarily, simply set the `open-pull-request-limit` to `0`.


luckysori (approved):
Just the PR I was looking for!

PR #464, opened 2022-11-30
feat: Add feedback button

### LINKED ISSUE #105: Share log messages
Dependent on #103 and #83.

We need a way to share log files from the TenTenOne App with us. using FLog we can make use of the `FLog.exportLogs()`

### LINKED ISSUE #414: Add send feedback button from within the app
The user should be able to send feedback to the app.

**Acceptance Criterias**

- [x] The button click should open an email template on the users phone
- [x] The logs should be automatically exported to file (zipped?) and attached to the email

holzeis:
I found this pretty cool feedback package (https://pub.dev/packages/feedback), that will allow the user to draw on the screen, add a feedback and send his logs via email.

I guess we we will not use that on a major release, but to get feedback on a beta version, it's pretty useful.

Works on Android and ios (not couldn't test on ios, since the simulator does not come with an email client). Updating the text "What's wrong" is kind of possible, but really ugly through translations. I tried it, but stopped again as it wasn't worth it. If you are very much bothered about it feel free to create a follow up PR 🙈 .

I struggled a bit with where to put that button. I could think of the following.

1. always visible: but where, it shouldn't distract too much.
2. in the drawer: i guess pretty hidden and the user might not find it.
3. on the wallet dashboard: visible at the beginning but moves out of sight the more the app is used. The user should still know that it's there though.

resolves #105 
resolves #414 

feedback

### DIFF
--- a/lib/wallet/wallet_dashboard.dart
+++ b/lib/wallet/wallet_dashboard.dart
@@ -1,5 +1,10 @@
+import 'dart:io';
+import 'dart:typed_data';
+
 import 'package:f_logs/f_logs.dart';
+import 'package:feedback/feedback.dart';
 import 'package:flutter/material.dart';
+import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
 import 'package:go_router/go_router.dart';
 import 'package:ten_ten_one/balance.dart';
@@ -13,6 +18,7 @@ import 'package:ten_ten_one/wallet/fund_wallet_on_chain.dart';
 import 'package:ten_ten_one/wallet/payment_history_list_item.dart';
 import 'package:ten_ten_one/wallet/seed.dart';
 import 'package:ten_ten_one/wallet/service_card.dart';
+import 'package:flutter_email_sender/flutter_email_sender.dart';
 
 import 'package:ten_ten_one/menu.dart';
 import 'package:ten_ten_one/app_bar_with_balance.dart';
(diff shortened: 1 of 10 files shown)


lib/wallet/wallet_dashboard.dart
```diff
@@ -92,6 +98,31 @@ class _WalletDashboardState extends State<WalletDashboard> {
...
+        child: const Text('Provide feedback'),
+        onPressed: () {
+          BetterFeedback.of(context).show((feedback) async {
+            final screenshotFilePath = await writeImageToStorage(feedback.screenshot);
+            final logs = await FLog.exportLogs();
+
+            final Email email = Email(
+              body: feedback.text,
+              subject: '10101 Feedback',
+              recipients: ['richard@coblox.tech'],
```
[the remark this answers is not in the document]
bonomat: I've created `10101-devs@coblox.tech` for now. We don't have emails for 10101.finance yet.

bonomat, line 135:
```suggestion
              recipients: ['10101-devs@coblox.tech'],
```

bonomat (approved):
Awesome find. I would also leave this plugin for the major release as we want to be able to collect feedback for quite a while :)

Re positioning, I would have put it into the menu/drawer.

luckysori (approved):
Awesome find! I would also put it in the drawer.

lib/wallet/wallet_dashboard.dart
```diff
@@ -92,6 +99,49 @@ class _WalletDashboardState extends State<WalletDashboard> {
     );
 
     widgets.add(paymentHistoryList);
+    if (Platform.isAndroid || Platform.isIOS || true) {
```
klochowicz, line 102:
```suggestion
```
no need for this `if` anymore, if it always evaluates to `true` 😉

klochowicz (approved):
great find!

klochowicz:
I will test it on iOS right now.

da-kami:
> Awesome find! I would also put it in the drawer.

Given that the number one purpose for this release is to collect feedback I'd definitely not hide the feedback button in the drawer. I think It's good on the dashboard because it is yet another action that we would expect the user to do. The look and feel being different than the action cards is good too from my perspective. I think it is good the way it is :)

da-kami (approved):
Awesome Flutter plugin! 
Makes me consider to become a sponsor :D

da-kami:
> > > Awesome find! I would also put it in the drawer.
> > 
> > 
> > Given that the number one purpose for this release is to collect feedback I'd definitely not hide the feedback button in the drawer. I think It's good on the dashboard because it is yet another action that we would expect the user to do. The look and feel being different than the action cards is good too from my perspective. I think it is good the way it is :)
> 
> I think we could do both. you would like to give feedback from other places, and dashboard is pretty haphazard.

True! I'm in favor for adding it to the drawer menu in addition to keeping it on the Dashboard :)

lib/wallet/wallet_dashboard.dart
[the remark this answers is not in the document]
holzeis: thanks

lib/wallet/wallet_dashboard.dart
holzeis: ups.. I added true only for testing 😅 good catch 👍

holzeis:
I added it to the drawer. It makes a lot of sense there since the drawer is always available and by using `Navigator.pop()` you will be able to give feedback directly at the page you are currently at.

luckysori (approved):

PR #467, opened 2022-11-30
fix: Navigate back to home on do it later

holzeis:
I guess "do it later" means going back to the home screen?


bonomat (approved):
Thanks for fixing it.

PR #468, opened 2022-11-30
fix: Hide close channel on empty lightning balance

### LINKED ISSUE #465: Close channel visible even if no channel is open
I can close a channel even if no channel is open. It would be nice if that option could be at least disabled.

Image

holzeis:
Not the nicest fix, but small and functional. Once we have a better depiction of the channel on the frontend we should adapt that.

resolves #465


bonomat (approved):
I think that fix is good enough for now.

luckysori (approved):
LGTM!

klochowicz (approved):

PR #469, opened 2022-11-30
chore: store wallet stuff inside a sub-dir

bonomat:
That's a bit cleaner in the directory

### DIFF
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -100,7 +100,7 @@ impl Wallet {
         )?;
 
         // Create a database (using default sled type) to store wallet data
-        let db = bdk::sled::open(data_dir.clone())?;
+        let db = bdk::sled::open(data_dir.join("wallet"))?;
         let db = db.open_tree(wallet_name)?;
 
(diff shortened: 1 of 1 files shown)


rust/src/wallet.rs
```diff
@@ -100,7 +100,7 @@ impl Wallet {
         )?;
 
         // Create a database (using default sled type) to store wallet data
-        let db = bdk::sled::open(data_dir.clone())?;
+        let db = bdk::sled::open(data_dir.join("wallet").clone())?;
```
luckysori, line 103: Is this a breaking change for the maker?

bonomat: yes, we'd need to move the seed files. Let's not do it then for now :)

bonomat:
Not worth the hassle

rust/src/wallet.rs
bonomat: actually no :D

bonomat:
> Not worth the hassle

Actually no hassle at all, there is nothing to do. Worst case we have the wallet db twice if we don't clean up.

luckysori (approved):

PR #471, opened 2022-11-30
chore: Remove spammy log message

klochowicz:
When this variable was not set (normal case for production),
it would print out 4 times every 20s announcing that it defaults to stable.

No need for logging this, the behaviour is well documented.


klochowicz:
screenshot of the behaviour without this patch:
image

luckysori (approved):
Oh lol, I expected this to be printed once. Oops!

Issue #475, opened 2022-12-01
Collab close transaction uses wrong amounts


bonomat:
it looks like that we use the wrong amounts when collaboratively closing: 

how to reproduce

1. open channel
2. open position
3. close position (and make a loss or profit)
4. close channel
5. get initial funding and win!

Step 5 is wrong obviously :)

luckysori:
Fixed with https://github.com/lightningdevkit/rust-lightning/commit/10b69c99ea13c6c2c55c74a7497764a6303b9d34, I'll create a PR shortly.

Issue #461, opened 2022-11-30
Closing channel requires 1 confirmation before funds are returned


holzeis:
Observed while debugging #440.

Is this delay intended?

---

EDIT: Used to be 6 confirmations, now it's 1 (maybe even 0 and I'm just impatient, but it's definitely not instant as the 2 transactions must be published and found in the mempool or on-chain) thanks to https://github.com/lightningdevkit/rust-lightning/commit/86531aa2ecdc228ae3788fda7332511c08870a69.

As @da-kami pointed out, a reasonable UX must be ensured regardless. The user needs to know that we are waiting for something to happen before the Lightning and Bitcoin balances are updated.

bonomat:
It looks like this is a hard requirement in the ldk: 
https://github.com/itchysats/rust-lightning/blob/f4f1093edc506314484c8d52a40dd155e692394b/lightning/src/chain/channelmonitor.rs#L397-L401

The field `ANTI_REORG_DELAY` is hardcoded to 6, letting us spend this output only after 6 confirmation

da-kami:
Interesting - do we see it as `pending` balance or not at all?

bonomat:
> Interesting - do we see it as `pending` balance or not at all?

not at all until after the 6 blocks

da-kami:
> > Interesting - do we see it as `pending` balance or not at all?
> 
> not at all until after the 6 blocks

That sucks, that's definitely not a nice UX :/

luckysori:
I've updated LDK to deal with this: https://github.com/lightningdevkit/rust-lightning/commit/86531aa2ecdc228ae3788fda7332511c08870a69. 10101 PR incoming.

PR #478, opened 2022-12-01
Remove feature to close channel

bonomat:
Due to time constraints I propose to remove this feature for the time being:
 
This postpones solving these problems (and potentially more) 
- https://github.com/itchysats/10101/issues/461: we have to wait for 6 confirmations before we even see the balance being updated
- https://github.com/itchysats/10101/issues/475: after closing a cfd the amount we receive should be different, however, this seems not work.

### DIFF
--- a/lib/wallet/wallet_lightning.dart
+++ b/lib/wallet/wallet_lightning.dart
@@ -70,21 +70,6 @@ class WalletLightning extends StatelessWidget {
       )
     ];
 
-    // close channel should only be possible if a channel is opened. the lightning balance is not
-    // directly indicating that a channel is open, but on the other hand no lightning balance can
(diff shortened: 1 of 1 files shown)


da-kami (approved):
I'm unhappy to have to do so, but I agree with @bonomat that the problems recorded in https://github.com/itchysats/10101/issues/461 and https://github.com/itchysats/10101/issues/475 are hard to "fix" within the timeframe we have left.

luckysori (commented):
I think we can fix the problems!

bonomat:
> I think we can fix the problems!

Looks like you did fix one of them :)

luckysori:
I've converted this to draft since it's tentative.

bonomat:
closing this for now as we might have a different solution

Issue #477, opened 2022-12-01
The maker's funding amout is not ignored on the taker's history once the channel is closed


da-kami:
Once the channel is closed we cannot query for the funding transaction to subtract the maker's funding amount.
We have to additionally save the funding-txid (once we use it for the first time) and then use that instead of extracting it from the channel.

This effects collaborative as well as non-collaborative close.

PR #479, opened 2022-12-01
Persist the channel opening `txid`

da-kami:
fixes https://github.com/itchysats/10101/issues/477

This enables us to properly ignore the maker's funding even when the channel was closed already. We save the `open_channel_txid` upon the first time we are able to extract it from the channel. Once persisted we will always load it from the database for consecutive calls to sync the payment history.


luckysori (approved):
Thanks for looking into this!

PR #480, opened 2022-12-01
chore: Ensure empty Strings are not passed into library

klochowicz:
Clearly state the contract of the API, ie. not supporting empty values.

By being explicit that we don't support empty strings, we make it easier for
ourselves to notice such errors.

lack of this made me hunt for a bug in my code way longer than expected :(

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -53,6 +53,7 @@ impl Address {
 
 #[tokio::main(flavor = "current_thread")]
 pub async fn init_db(app_dir: String) -> Result<()> {
+    anyhow::ensure!(!app_dir.is_empty(), "app_dir must not be empty");
     let network = config::network();
     db::init_db(
         &Path::new(app_dir.as_str())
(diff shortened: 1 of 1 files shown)


rust/src/api.rs
```diff
@@ -73,6 +74,7 @@ pub async fn test_db_connection() -> Result<()> {
 }
 
 pub fn init_wallet(path: String) -> Result<()> {
+    anyhow::ensure!(!path.is_empty(), "path  must not be empty");
```
luckysori, line 77:
```suggestion
    anyhow::ensure!(!path.is_empty(), "path must not be empty");
```

rust/src/api.rs
```diff
@@ -184,6 +188,7 @@ pub async fn get_offer() -> Result<Option<Offer>> {
 
 #[tokio::main(flavor = "current_thread")]
 pub async fn call_faucet(address: String) -> Result<String> {
+    anyhow::ensure!(!address.is_empty(), "received empty address");
```
luckysori, line 191: 🙃 Might be hard to tell this one apart from the previous one if we use the same error message.

I like the pattern that you used for the last error: `Cannot do X because of empty Y`.

luckysori (approved):

da-kami (approved):

rust/src/api.rs
klochowicz: I believe they have the line number attached, but I concur that it will be easier to distinguish if I added a bit more info. added :)

PR #481, opened 2022-12-01
Restrict tooltip to size of Bitcoin balance

da-kami:
The tooltip was created around the whole `Row`, which could cause trouble with opening the menu drawer in the `Bitcoin` wallet screen. The menu drawer was not properly clickable because the tooltip partly overlayed it.

Fixed by moving restricting the size of the `Row` to min (so it does not auto-expand) and moving the `Center` to the outside (it should not have been part of the `Balance` in the first place, the caller should decide how to layout the balance).

### DIFF
--- a/lib/app_bar_with_balance.dart
+++ b/lib/app_bar_with_balance.dart
@@ -34,7 +34,7 @@ class AppBarWithBalance extends StatelessWidget {
           ),
           Padding(
             padding: const EdgeInsets.only(left: 20.0, right: 20.0, top: 20.0, bottom: 10.0),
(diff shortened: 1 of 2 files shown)


luckysori (approved):
Wow, I never encountered this. But I was just able to reproduce it on `main` following the steps you described. Sorry about that!

da-kami:
> Wow, I never encountered this. But I was just able to reproduce it on `main` following the steps you described. Sorry about that!

No worries 😊
It's partly my fault, because the `Balance` should have not contained the `Center` - this was part of the problem - and I should have probably restrained the size of the `Row` in the first place. Nice learnings :)

PR #482, opened 2022-12-01
Avoid 0-conf channels + closing channels a bit faster

luckysori:


rust/src/lightning.rs
```diff
@@ -152,28 +152,15 @@ pub async fn open_channel(
     channel_amount_sat: u64,
     initial_send_amount_sats: u64,
 ) -> Result<()> {
-    let config = UserConfig {
-        channel_handshake_limits: ChannelHandshakeLimits {
-            // lnd's max to_self_delay is 2016, so we want to be compatible.
-            their_to_self_delay: 2016,
```
da-kami, line 158: 🤔 Was this line removed on purpose? I feel it is unrelated to the changes described.

da-kami (approved):

rust/src/lightning.rs
bonomat: Not really, but not needed for our use at the moment.

luckysori:
I'll drop the second commit, because it'll conflict with https://github.com/itchysats/10101/pull/489 and it is included in it.

rust/src/lightning.rs
bonomat: We will notice once we try 😂

Issue #433, opened 2022-11-29
The connection management should act on the connection instead of the alive endpoint


holzeis:
Currently we are miss-using the makers alive endpoint to check if the connection is still alive. This is not ideal as the connection could be lost, even if the maker is still alive.

The status of the connection should be implied directly from the connection. The peer manager seems to be able to process a disconnect event.

PR #483, opened 2022-12-01
Try to reconnect to maker if channel is not usable

luckysori:
According to the `rust-lightning` docs if the channel is not ready, it could be because the connection with the maker was lost.


holzeis:
I guess this will also resolve #433

holzeis (approved):

PR #486, opened 2022-12-01
fix: Center the dashboard tiles

klochowicz:
It's not a ListView anymore, but given the amount of the tiles there's no way to
overflow it on any mobile screen (tested on iPhone SE3)

### DIFF
--- a/lib/wallet/wallet_dashboard.dart
+++ b/lib/wallet/wallet_dashboard.dart
@@ -124,25 +124,22 @@ class ServiceNavigation extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return SizedBox(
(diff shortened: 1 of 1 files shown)


luckysori:
For reference, before:

image

After:

image

luckysori (approved):

holzeis (approved):
I likeed the non centered version better, though. 🙈

klochowicz:
> I liked the non centered version better, though. 🙈

that's interesting! and indicates that perhaps we haven't got to the ideal solution yet!

klochowicz:
> For reference, before:
> 
> image
> 
> After:
> 
> image

thank you for providing this 🙏

PR #487, opened 2022-12-01
Set the correct lightning port for `testnet` + `main`

da-kami:


holzeis (approved):
It hurts me to see that config in code, but ok, I guess?

luckysori:
> It hurts a little bit to see that config in code, but ok, I guess?

I just want to mention that this is just following a pattern that we introduced a while back!

PR #489, opened 2022-12-01
Fix collaborative channel close amounts

luckysori:
Fixes https://github.com/itchysats/10101/issues/475.

They were wrong because of a bug in our fork of `rust-lightning`: the channel amounts were not being updated after the custom output was removed.

We were already constructing the correct commitment transaction and because we only ever tested using force-close it looked like everything was working as intended!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1298,7 +1298,7 @@ dependencies = [
 [[package]]
 name = "lightning"
 version = "0.0.112"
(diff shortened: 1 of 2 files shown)


bonomat (approved):
Awesome find. Thanks Lucas.

da-kami (approved):
❤️

PR #495, opened 2022-12-01
fix: Allow manual input of lightning invoice

holzeis:
Fixes a minor regression that invoices can't be entered manually anymore. Also when looking into this bug, I figured it's probably nicer to have the QR code scanning icon next to the field. That way we could make it a reusable widget.


klochowicz (approved):
thanks, looks good!

Issue #485, opened 2022-12-01
Disable channel actions until it is usable


bonomat:
We should not allow any actions on the channel (CFD, payments) until it is actually usable.

bonomat:
I've started with this and added the API but it's not being used yet in the frontend. Feel free to pick it up and continue: 
https://github.com/itchysats/10101/compare/feat/usable-channel?expand=1

bonomat:
https://github.com/itchysats/10101 adds already the functionality we need. Next step is to simply disable the buttons in the frontend.

PR #496, opened 2022-12-01
feat: Wait for channel to become available

holzeis:
Before that PR we where only depending on a timeout and a lightning balance. However, this led to several error messages as a lightning balance does not mean the channel is usable.

The PR introduces a sync of the channel state every 5 seconds representing the following 3 states:

1. Unavailable: No channel exists
2. Establishing: Channel exists but is not usable yet
3. Available: Channel exists and is usable

ebe4fb219a61b1802e673723916a77b8bad79ad4 fixes a regression after merging https://github.com/itchysats/10101/pull/483: we can not assume there is always a channel available. The taker was trying to reconnect constantly if no channel was available or if it was in creation. It's hard to distinguish between a good and a bad usable channel (as it might be in creation) - I am using the `short_channel_id` now to distinguish if the channel is actually in a bad state or everything is fine.

### DIFF
--- a/lib/wallet/wallet_dashboard.dart
+++ b/lib/wallet/wallet_dashboard.dart
@@ -40,7 +40,6 @@ class _WalletDashboardState extends State<WalletDashboard> {
   Widget build(BuildContext context) {
     final seedBackupModel = context.watch<SeedBackupModel>();
     final bitcoinBalance = context.watch<BitcoinBalance>();
-    final lightningBalance = context.watch<LightningBalance>();
     final paymentHistory = context.watch<PaymentHistory>();
     final channel = context.watch<ChannelChangeNotifier>();
 
@@ -65,7 +64,7 @@ class _WalletDashboardState extends State<WalletDashboard> {
           icon: const Icon(Icons.link))));
     }
 
-    if (bitcoinBalance.total().asSats != 0 && lightningBalance.amount.asSats == 0) {
+    if (bitcoinBalance.total().asSats > 0 && !(channel.isAvailable() || channel.isInitialising())) {
       widgets.add(ActionCard(CardDetails(
         route: OpenChannel.route,
         title: "Open Channel",
--- a/rust/src/connection.rs
+++ b/rust/src/connection.rs
@@ -7,9 +7,7 @@ use anyhow::Result;
 pub async fn keep_alive() -> Result<()> {
     let mut connected = false;
     loop {
-        let is_connection_lost_maybe = !is_first_channel_usable()?;
-
-        if !connected || is_connection_lost_maybe {
+        if !connected || !is_first_channel_usable() {
             connected = connect().await?;
         }
 
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -521,13 +521,11 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()> 
 /// If the first channel is not usable, it might be because we've lost
 /// the connection with the 10101 maker, according to the
 /// `rust-lightning` logs.
-pub fn is_first_channel_usable() -> Result<bool> {
-    let is_usable = match get_first_channel_details() {
+pub fn is_first_channel_usable() -> bool {
+    match get_first_channel_details() {
         Some(channel_details) => channel_details.is_usable,
         None => false,
-    };
-
-    Ok(is_usable)
+    }
 }
 
 pub fn get_first_channel_details() -> Option<ChannelDetails> {
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -544,7 +544,7 @@ pub async fn connect() -> Result<()> {
         lightning.peer_manager.clone()
     };
     let peer_info = maker_peer_info();
-    tracing::debug!("Connection with {peer_info}");
+    tracing::debug!("Connecting with {peer_info}");
     lightning::connect_peer_if_necessary(&peer_info, peer_manager).await?;
 
     Ok(())
--- a/rust/src/wallet.rs
+++ b/rust/src/wallet.rs
@@ -14,6 +14,7 @@ use crate::lightning::NodeInfo;
 use crate::lightning::PeerInfo;
 use crate::seed::Bip39Seed;
 use ::lightning::chain::chaininterface::ConfirmationTarget;
+use ::lightning::ln::channelmanager::ChannelDetails;
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
@@ -521,18 +522,21 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()> 
 /// the connection with the 10101 maker, according to the
 /// `rust-lightning` logs.
 pub fn is_first_channel_usable() -> Result<bool> {
-    let channel_manager = {
-        let lightning = &get_wallet()?.lightning;
-
-        lightning.channel_manager.clone()
-    };
-
-    let is_usable = match channel_manager.list_channels().first() {
+    let is_usable = match get_first_channel_details() {
         Some(channel_details) => channel_details.is_usable,
-        None => return Ok(false),
+        None => false,
     };
 
     Ok(is_usable)
+}
+
+pub fn get_first_channel_details() -> Option<ChannelDetails> {
+    let channel_manager = match &get_wallet() {
+        Ok(wallet) => Some(wallet.lightning.channel_manager.clone()),
+        Err(_) => None,
+    }?;
+
+    channel_manager.list_channels().first().cloned()
 }
 
 pub async fn connect() -> Result<()> {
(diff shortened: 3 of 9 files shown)


rust/src/connection.rs
```diff
@@ -7,7 +7,11 @@ use anyhow::Result;
 pub async fn keep_alive() -> Result<()> {
     let mut connected = false;
     loop {
-        let is_connection_lost_maybe = !is_first_channel_usable()?;
+        let channel_details = wallet::get_first_channel_details();
+
+        let is_connection_lost_maybe = channel_details.is_some()
+            && !is_first_channel_usable()
+            && channel_details.unwrap().short_channel_id.is_some();
```
bonomat, line 14: I would not unwrap here but use `unwrap_or_default` which evaluates to false.

bonomat (approved):
Thanks. I think this helps with https://github.com/itchysats/10101/issues/485

lib/wallet/wallet_dashboard.dart
```diff
@@ -65,7 +64,7 @@ class _WalletDashboardState extends State<WalletDashboard> {
           icon: const Icon(Icons.link))));
     }
 
-    if (bitcoinBalance.total().asSats != 0 && lightningBalance.amount.asSats == 0) {
+    if ((bitcoinBalance.total().asSats > 0 && !channel.isAvailable()) || channel.isInitialising()) {
```
da-kami, line 67: ❓ `|| channel.isInitialising()` was this condition added so we see the card with the spinner?

bonomat:
I'm taking over this last bit :)

rust/src/wallet.rs
```diff
@@ -521,20 +522,23 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()>
...
+    let is_usable = match get_first_channel_details() {
         Some(channel_details) => channel_details.is_usable,
-        None => return Ok(false),
+        None => false,
     };
 
     Ok(is_usable)
 }
 
+pub fn get_first_channel_details() -> Option<ChannelDetails> {
```
da-kami, line 531: Noice! This one will come in handy in multiple places

rust/src/connection.rs
bonomat: Second thought: `            && channel_details.unwrap().short_channel_id.is_some();`
should be redundant as `is_first_channel_usable`, i.e. `channel.is_usable` is only true if the channel is confirmed.

da-kami, line 14: Let me try to understand this:

`is_connection_lost_maybe` describes that the connection is (maybe) *lost*, and it is lost if we have a channel (`channel_details.is_some()`) and this channel is not usable (`!is_first_channel_usable()`) and we can retrieve the short channel id from said channel (`channel_details.unwrap().short_channel_id.is_some()`)...?

I think the last condition does not make sense - I think this should be:

```
 let is_connection_lost_maybe = channel_details.is_some()
            && (!is_first_channel_usable()
            || !channel_details.unwrap().short_channel_id.is_some());
```

lib/wallet/wallet_dashboard.dart
klochowicz, line 67: was your intention to have the "Open Channel' button available when the channel is already initialising?

In case it wasn't, then I've reworked it slightly to disable button when it's initialising too:

```suggestion
    if (bitcoinBalance.total().asSats > 0 && !(channel.isAvailable() || channel.isInitialising())) {
```

rust/src/connection.rs
bonomat: Third thought: I think this does not work as you want.

if we do have a channel which is currently being confirmed `is_first_channel_usable=false` and `channel_details.unwrap().short_channel_id.is_some()=false`  then we would constantly reconnect.

rust/src/wallet.rs
```diff
@@ -520,19 +521,20 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()>
...
 /// `rust-lightning` logs.
-pub fn is_first_channel_usable() -> Result<bool> {
-    let channel_manager = {
-        let lightning = &get_wallet()?.lightning;
-
-        lightning.channel_manager.clone()
-    };
-
-    let is_usable = match channel_manager.list_channels().first() {
+pub fn is_first_channel_usable() -> bool {
```
klochowicz, line 524: note: given that the channels might appear in any order, we will need to rework this when we decide to have more than one channel open (e.g. simultaneous cfd and lightning functionality)

rust/src/connection.rs
bonomat: Maybe I'm missing something, but do we even need this logic. 
In `connect_peer_if_necessary` we check if the peer is known and hence we have a connection: 

https://github.com/itchysats/10101/blob/576e51a8cff6668ac1cf57f8f32584d0d8c0b124/rust/src/lightning.rs#L637-L641

bonomat: cc @da-kami / @klochowicz

rust/src/wallet.rs
```diff
@@ -520,19 +521,20 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()>
...
-    };
+        None => false,
+    }
+}
+
+pub fn get_first_channel_details() -> Option<ChannelDetails> {
+    let channel_manager = match &get_wallet() {
+        Ok(wallet) => Some(wallet.lightning.channel_manager.clone()),
+        Err(_) => None,
+    }?;
```
klochowicz, line 535: nit: see below (gh didn't allow me doing it in one because the empty line on line 536 was there before 😂)
```suggestion

```

rust/src/wallet.rs
```diff
@@ -520,19 +521,20 @@ pub async fn open_channel(peer_info: PeerInfo, taker_amount: u64) -> Result<()>
...
 
-    Ok(is_usable)
+    channel_manager.list_channels().first().cloned()
```
klochowicz, line 537: nit:
I think we actually have a helper function `get_channel_manager()`, which returns a `Result`. With this, we can use `ok()` to convert it to an `Option` easily!

```suggestion
    get_channel_manager().ok().map(|cm| cm.list_channels().first().cloned())
```

rust/src/connection.rs
da-kami: My understanding:

We used `connect_peer_if_necessary` and it did not work stable, i.e. we encountered problems with the connection when e.g. opening a channel before. I don't know if we can rely on the `peer_manager` to decide if the connection is healthy - problems that we encountered in the past would point to no.

That's why additional logic is being added to decide if the connection is still considered healthy.

bonomat: basically, I would change this logic to [`477e8e2` (#496)](https://github.com/itchysats/10101/pull/496/commits/477e8e2bb07f9b386492e2f69bc23916ddcd3fd1)

klochowicz (approved):

lib/wallet/wallet_dashboard.dart
bonomat: Good catch!

rust/src/wallet.rs
luckysori, line 524: 🔧 This value could be derived from `get_channel_state` so that the logic is not effectively duplicated.

luckysori (approved):

bonomat:
@klochowicz / @luckysori / @da-kami : please have another look at my fixup: [`477e8e2` (#496)](https://github.com/itchysats/10101/pull/496/commits/477e8e2bb07f9b386492e2f69bc23916ddcd3fd1). I don't think the logic is needed: 

- if there is no channel, 
  - `is_first_channel_usable=false`, try to reconnect
- if there is a channel
	- `is_first_channel_usable=true`: we are connected, no need to reconnect
	- `is_first_channel_usable=false`: we are not conncted, reconnect
	
`is_first_channel_usable` checks if `channel_state=ChannelState::PeerDisconnected`. So no need to have another check

rust/src/connection.rs
```diff
@@ -7,13 +7,7 @@ use anyhow::Result;
...
     let mut connected = false;
     loop {
-        let channel_details = wallet::get_first_channel_details();
-
-        let is_connection_lost_maybe = channel_details.is_some()
-            && !is_first_channel_usable()
-            && channel_details.unwrap().short_channel_id.is_some();
-
-        if !connected || is_connection_lost_maybe {
+        if !connected || !is_first_channel_usable() {
```
luckysori, line 10: If I understand it correctly, `is_first_channel_usable` already checks that the channel exists, so we don't need the first guard.

Also, we think that the channel will only be usable if the `short_channel_id` is present.

Makes sense to me!

luckysori (approved):

bonomat:
Rebased and going ahead to merge this.

bors r+

lib/wallet/wallet_dashboard.dart
holzeis: yes. It can take quite some time, before the channel is actually ready, even if the app already shows a lightning balance. I've added this to indicate to the user, the channel is not ready yet. It's not the nicest design, but it's better than a warning message when trying to open a cfd.

holzeis: @bonomat  @klochowicz 

well it was intentional! I wanted to show the card when it is initialising, giving the user feedback that the channel is still in establishment. The card is disabled if the channel is initialising

PR #500, opened 2022-12-02
feat: Highlight active page in the sidebar menu

klochowicz:
Makes it much easier to orient oneself in the app.

Sadly, ExpandedTile highlight ("Stacking") cannot be tweaked at all (always selected when expanded) , so I resorted to
also highlight the background of the active tile to make it stand out more.

image


klochowicz:
the problem itself was mentioned a few times, but we never managed to put it in a ticket :/

bonomat (approved):

Issue #411, opened 2022-11-28
Add on-boarding tour


holzeis:
Upon the first start of the app the use should get a tour through the app explaining whats what and how to CFD!

bonomat:
We should have a look into https://bitcoin.design/guide/designing-products/common-user-flows/#software-onboarding
They have some nice on-boarding flows we could copy.

da-kami:
Would an additional onboarding tour on first start be enough (e.g. using https://pub.dev/packages/onboarding) or wold you add onboarding to different parts of the application?

I think in the long run we need *some* onboarding for different parts, i.e. explain what the wallet backup means, have simple onboarding into the flow of fund on-chain -> open channel -> trade, some onboarding into CFD trading.

For now I propose to just add some initial onboarding using a package like https://pub.dev/packages/onboarding where we explain that:

1. There is an on-chain and a lightning wallet
2. The user has to fund the on-chain wallet to open a channel
3. Once the channel was opened the user can trade.

bonomat:
I would just go for a single on-boarding screen for the whole app, welcome the user and explain what this app is about. 

The plugin you propose seems reasonable to me :)

da-kami:
OK! I can give that a go. I'll assign myself.

PR #503, opened 2022-12-02
Add onboarding tour on first startup

da-kami:
resolves https://github.com/itchysats/10101/issues/411

On first startup we show an onboarding tour including three screens:

1. one App all things Bitcoin: Welcome screen introducing 10101
2. your Keys, your Bitcoin: Explain that there is a Bitcoin and Lightning wallet that are created on first startup
3. Trade P2P non-custodial: List the three main steps of deposit, open channel, trade. Tell the user that the app will guide the user through these steps.

The onboarding tour is currently only shown on first startup, we persist a bool after that in the shared preferences. When the app is uninstalled and re-installed the onboarding tour is shown again (give that the app data was removed).

---

Notes: 

The `onboarding` flutter package is actually somewhat annoying, if I would do this again I'd probably code this myself, but I only noticed some annoyances when I was already half way through...

E.g. I think it would be nice to be able to click the three dots to navigate, but it does not have that (I think...?).
Changing the colors and theming is quite annoying, and I could not figure out how to change some...

When there is only little text / a small image the page (that one swipes) does not expand properly, so one cannot swipe properly. We might run into this problem on bigger screens. I tried for some time to overcome this, but it resulted in other problems that were annoying to fix without changing the library :/ 

---

<details>
<summary> 👉 Unfold here to see the videos with the initial design. This includes a video where one see the second startup without the tour.</summary>

Recording of the tour after a fresh install and initial startup:

Close the app and start it again, the tour is not shown (not easy to close an app on iPhone SE...):

</details>


da-kami:
iPhone 14 pro:

Screenshot 2022-12-02 at 4 46 16 pm

Desktop:

Screenshot 2022-12-02 at 4 45 56 pm

Note: Desktop also defaults to wiping, which is stupid 😅

lib/menu.dart
```diff
@@ -77,6 +78,14 @@ class Menu extends StatelessWidget {
         leading: const Icon(Icons.settings),
         onTap: () {},
       ),
+      ListTile(
+        title: const Text("Onboarding"),
+        leading: const Icon(Icons.help),
```
klochowicz, line 83: after rebasing on `main`, you can follow the pattern to make it selectable

lib/onboarding_tour.dart
```diff
@@ -0,0 +1,190 @@
...
+  late int index;
+
+  final onboardingPagesList = [
+    PageModel(
+      widget: SingleChildScrollView(
+        controller: ScrollController(),
+        child: OnboardingPage(
+            title: "WELCOME TO 10101",
+            description:
+                "The one-stop Bitcoin app with non-custodial trading over Lightning. 10101 is all about keeping your keys where they belong: with you.",
```
klochowicz, line 26: nit: It would be nice to suggest to the user that it's TenTenOne to clear any ambiguity

```suggestion
                "The one-stop Bitcoin app with non-custodial trading over Lightning. 10101 (TenTenOne) is all about keeping your keys where they belong: with you.",
```

klochowicz (approved):
LGTM! thanks for tackling this!

lib/menu.dart
da-kami: Ah, crap, that's not supposed to be there, I committed that accidentally.

lib/onboarding_tour.dart
da-kami: I can write it as a word on the first screen, but I am not convinced that the onboarding has to mention how to pronounce the name 😅

lib/menu.dart
da-kami: I removed it in another fixup - if you think it's useful to keep it in the menu we can also keep it.

klochowicz: I think it is useful indeed! if someone wants to trigger it, they can. makes the app feel a bit more complete.

lib/onboarding_tour.dart
klochowicz: to me this is part of onboarding - there's no other spot to teach the users :P

klochowicz:
feedback after trying it:
- it would be good to use our 10101 logo on the first screen, I was slightly surprised and disturbed with some different logo there (but this might be subjective
- the bottom bar could be smaller for the things to fit in better - this might be one of the awkwardness of the flutter package. no dramas if it stays this way. 

overall, really cool!

da-kami:
I am doing another iteration on making the look and feel better. I played with the colors a bit more:

Screenshot 2022-12-02 at 5 19 50 pm

I will also try to fix the expanding of the column, but so far I was unable to achieve that. (I tried quite a bit 😅)

da-kami:
> * it would be good to use our 10101 logo on the first screen, I was slightly surprised and disturbed with some different logo there (but this might be subjective

I thought the "swiss-army knife of lightning wallets" was a fun idea... but maybe it's not as good as I thought. I can re-design this screen. Any ideas?

klochowicz:
> > * it would be good to use our 10101 logo on the first screen, I was slightly surprised and disturbed with some different logo there (but this might be subjective
> 
> I thought the "swiss-army knife of lightning wallets" was a fun idea... but maybe it's not as good as I thought. I can re-design the welcome screen. Any ideas?

it is fun! I got perhaps surprised because I missed the splash screen when testing (didn't look at the screen as it was building). with normal startup, it shouldn't be as confusing.

lib/menu.dart
da-kami: OK, I can add it in a separate commit and then we can decide :)

da-kami: > after rebasing on main, you can follow the pattern to make it selectable

This actually does not make much sense, because the onboarding tour is a separate set of screens that does not have a burger menu. I can set it to selected, but you would basically never see it 😅

bonomat (approved):
Looks nice.

PR #504, opened 2022-12-02
fix: re-org bug was fixed upstream

bonomat:
see PR here: https://github.com/itchysats/bdk-ldk/pull/8


bonomat:
already approved upsteram

bors r+

PR #509, opened 2022-12-02
Update on-chain balance immediately after channel closure

### LINKED ISSUE #461: Closing channel requires 1 confirmation before funds are returned

luckysori:
Fixes #461.

The coins are correctly moved to a pending state, like we do with any other on-chain transaction.

The user is also informed about what to expect when closing the channel.

It's important to note that if the user runs into a force-closure for some other reason, the coins will also go into a pending state and be reflected in the overall on-chain balance, but it will take _hundreds_ of confirmations for the coins to be confirmed.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,7 +219,7 @@ dependencies = [
 [[package]]
 name = "bdk-ldk"
 version = "0.1.0"
-source = "git+https://github.com/itchysats/bdk-ldk?rev=c66ad845391a67a80554962bab55a237775150df#c66ad845391a67a80554962bab55a237775150df"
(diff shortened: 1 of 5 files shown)


holzeis (approved):
LGTM 👍

bonomat (commented):
Very nice and simple solution :)

Issue #514, opened 2022-12-02
Can't see navigation menu


bonomat:
When coming onto this screen, the keyboard pops up by default and it is not possible to navigate to open positions . 

![image](https://user-images.githubusercontent.com/224613/205406121-cb3f8d2d-975c-498c-a377-714b1a20e0f7.png)

da-kami:
This might be a regression bug, on android I don't have a FAB at all if there is no channel. (I closed the keyboard and there is no FAB.) I don't think this screen should behave like this.

![image](https://user-images.githubusercontent.com/5557790/205520326-3f7ed30e-64f7-471f-9386-ea97b2e8c554.png)

bonomat:
> This might be a regression bug, on android I don't have a FAB at all if there is no channel. (I closed the keyboard and there is no FAB.) I don't think this screen should behave like this.

That's by design: if there is no action, there is no FAB

da-kami:
> That's by design: if there is no action, there is no FAB

OK, I was not aware that this was changed. Previously we kept the navigation to the subscreen (where `Confirm` was disabled) so the user can get a better feeling of the app even if the channel is not open yet.

da-kami:
> When coming onto this screen, the keyboard pops up by default and it is not possible to navigate to open positions .

One way to fix this is to turn the auto-open of the keyboard off. I;ll create a PR :)

PR #515, opened 2022-12-04
Do not auto-open the keyboard

da-kami:
#fixes https://github.com/itchysats/10101/issues/514

If we auto-open the keyboard the user is unable to navigate to other views (e.g. My CFDs) of the is screen.

Note: I don't think there is a better way to fix this without a bigger redesign of that screen - we could try to push the bottom menu up with the keyboard, but I think this will mess with the screen overall, because we don't have enough space to do that.

### DIFF
--- a/lib/cfd_trading/cfd_offer.dart
+++ b/lib/cfd_trading/cfd_offer.dart
@@ -176,7 +176,6 @@ class _CfdOfferState extends State<CfdOffer> {
                       keyboardType: TextInputType.number,
                       inputFormatters: [FilteringTextInputFormatter.digitsOnly],
                       style: const TextStyle(fontSize: 18.0),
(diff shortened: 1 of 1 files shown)


bonomat (approved):
That's a good start. I think the scroll view on this screen is messed up, hence you can't scroll down to see the bottom bar.

da-kami:
> That's a good start. I think the scroll view on this screen is messed up, hence you can't scroll down to see the bottom bar.

Usually the bootom bar is not part of the scrollable are of the screen. The bottom bar is in a fixed location, independent of the body being scrollable or not. But maybe one can configure the bottom bar to not be "fixed" - I have never tried that.

I think the better solution in the long run is to re-design the screen, and potentially the whole CFD trading navigation flow.

PR #517, opened 2022-12-04
Some small improvements

da-kami:


klochowicz (approved):

lib/wallet/open_channel.dart
```diff
@@ -87,7 +87,7 @@ class _OpenChannelState extends State<OpenChannel> {
                       crossAxisAlignment: CrossAxisAlignment.start,
                       children: [
                         Text(
-                            "Define the amount of bitcoin you want to lock into the Lightning channel. 10101 will double it for great inbound liquidity. Max channel size is currently $takerChannelAmount sats.",
+                            "Define the amount of bitcoin you want to lock into the Lightning channel. 10101 will double it for great inbound liquidity. The max amount you can use is currently $takerChannelAmount sats.",
```
luckysori, line 90: ❓ Isn't this text dynamic with the amount the user inputs? It still reads weird to me, because it seems to imply that the user cannot go over this number, but they literally can.

luckysori (approved):

lib/wallet/open_channel.dart
da-kami: Good point, I'll add the balance as validation constraint as well.

da-kami: https://github.com/itchysats/10101/pull/517/commits/0da990490f9b5c4f0aa24478082441ba5db3fcd4

Note: I think the balance does not *have* to be confirmed to open a channel, but getting the logic for pending *or* confirmed right might introduce regression bugs so I opted for `confirmed` right now.

da-kami: We could actually introduce form validation like for the CFD amount here, but I am not sure it should get priority

luckysori: Hmm, but I think we already achieve that with [line 37](https://github.com/itchysats/10101/pull/517/commits/7fba080297b362580597c80f1ee88144303dba24#diff-f66cc5c065055955fc377c471cc79795967d0fa0e2fdc56a0153e71ebeb747a1L37).

What I was referring to, is that the text changes with the value that the user inputs e.g.

```
The max amount you can use is currently 50 sats.
```

I think what we are actually trying to say here is that the _maximum_ outbound capacity that we allow is 500_000 sats (based on `OpenChannel.maxChannelAmount`). Or am I mistaken?

Otherwise it reads weird to me. The number changes as the user inputs the very same number just below.

https://user-images.githubusercontent.com/9418575/205525727-e26aff39-faea-459e-918d-9ef636adcb23.mp4

da-kami: Fair enough, I'll change the value that we display to be the balance, capped by the maximum amount

da-kami:
I came to the conclusion that with adapting our self-made form-validation the UX got worse and worse the more we restricted the validation. I came to the conclusion that we should introduce proper form validation and did so in https://github.com/itchysats/10101/pull/517/commits/92c885bcd6259631fff29a81cf4bcec68eca3ea6

da-kami:
Note that https://github.com/itchysats/10101/pull/517/commits/6195360f3daf0a9723cb527a019caddd2e78977f slightly changes the logic, because we allow channel open from a pending balance. @luckysori @bonomat I think this should be fine, but if we think this might cause trouble I can remove the commit as well. Please let me know

lib/wallet/open_channel.dart
```diff
@@ -25,131 +24,148 @@ class OpenChannel extends StatefulWidget {
...
+                                int intVal = int.parse(value.replaceAll(",", ""));
+
+                                if (intVal > maxTakerChannelAmount) {
+                                  return "Max channel amount of $maxTakerChannelAmount";
+                                }
+                                if (intVal < minValue) {
+                                  return "Min channel amount is $minValue";
+                                }
+                              } on Exception {
+                                return "Enter a number";
```
luckysori, line 130: 🙃 Not sure what the standard is, but this sounds a bit rude haha.

lib/wallet/open_channel.dart
```diff
@@ -25,131 +24,148 @@ class OpenChannel extends StatefulWidget {
...
+                                } on Exception {
+                                  takerChannelAmount = 0;
+                                }
+                              });
+
+                              _formKey.currentState!.validate();
+                            },
+                            validator: (value) {
+                              if (value == null) {
+                                return "Enter channel amount";
```
luckysori, line 117: 🤡 I find `channel amount` (or channel size) a bit vague. What the user is inputting is their outbound liquidity. Perhaps `Enter channel outbound liquidity`?

luckysori: Same issue with other usages of `amount` in this file.

lib/wallet/open_channel.dart
```diff
@@ -25,126 +24,149 @@ class OpenChannel extends StatefulWidget {
...
-                        "Hit the 'Open Channel' button and the channel will be created in a few moments.",
-                        style: TextStyle(color: Colors.grey))),
-                const SizedBox(height: 15),
-              ],
+
+                              try {
+                                int intVal = int.parse(value.replaceAll(",", ""));
+
+                                if (intVal > maxTakerChannelAmount) {
+                                  return "Max channel amount of $maxTakerChannelAmount";
```
da-kami, line 124: Should we state `insufficient balance` instead?

lib/wallet/open_channel.dart
```diff
@@ -43,8 +43,9 @@ class _OpenChannelState extends State<OpenChannel> {
   @override
   Widget build(BuildContext context) {
     final bitcoinBalance = context.watch<BitcoinBalance>();
-    final maxTakerChannelAmount =
-        bitcoinBalance.confirmed.asSats.clamp(0, OpenChannel.maxChannelAmount);
+    final maxTakerChannelAmount = bitcoinBalance.pendingExternal.asSats != 0
+        ? bitcoinBalance.pendingExternal.asSats.clamp(0, OpenChannel.maxChannelAmount)
+        : bitcoinBalance.confirmed.asSats.clamp(0, OpenChannel.maxChannelAmount);
```
luckysori, line 48: 🔧 What about `pendingInternal` (change outputs)? I think you _always_ want to use `confirmed + pendingInternal + pendingExternal` to determine the `maxTakerChannelAmount`.

Interestingly, we have a `total` method on `BitcoinBalance` which includes them, but it also includes `pendingChannelClose` which would **not** be safe to use here, as those UTXOs are not tracked by our on-chain wallet yet!

In conclusion, my suggestion would be:

```
final maxTakerChannelAmount = (bitcoinBalance.confirmed.asSats + bitcoinBalance.pendingInternal.asSats + bitcoinBalance.pendingExternal.asSats).clamp(0, OpenChannel.maxChannelAmount);
```

lib/wallet/open_channel.dart
luckysori: Not necessarily true, because we have the arbitrary constraint of 500,000 sats.

lib/wallet/open_channel.dart
da-kami: 😅 My if-or did not make any sense, of course one has to sum them up 🤦‍♂️

lib/wallet/open_channel.dart
da-kami: I changed it to `outbound liquidity` in most places.

PR #523, opened 2022-12-05
Clear expired lightning payments

klochowicz:
See commit details for more details.

screenshots
image

payment details:
image

### DIFF
--- a/rust/src/db.rs
+++ b/rust/src/db.rs
@@ -349,7 +349,7 @@ pub async fn clean_expired_payments() -> Result<()> {
         .collect::<Vec<PaymentInfo>>();
 
     for payment in newly_expired {
-        update_payment(&payment.hash, HTLCStatus::Failed, None).await?;
+        update_payment(&payment.hash, HTLCStatus::Expired, None).await?;
     }
     Ok(())
 }
@@ -462,7 +462,7 @@ mod tests {
         clean_expired_payments().await.unwrap();
         assert_eq!(
             load_payment(&hash).await.unwrap().status,
-            HTLCStatus::Failed
+            HTLCStatus::Expired
         );
     }
 }
--- a/rust/src/db.rs
+++ b/rust/src/db.rs
@@ -164,12 +164,13 @@ pub async fn insert_payment(payment: &PaymentInfo) -> Result<()> {
         amt_msat,
         created_timestamp,
         updated_timestamp,
+        expiry_timestamp,
     } = payment.clone();
 
     let query_result = sqlx::query(
         r#"
-        INSERT INTO payments (payment_hash, preimage, secret, flow, htlc_status, amount_msat, created, updated)
-        VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+        INSERT INTO payments (payment_hash, preimage, secret, flow, htlc_status, amount_msat, created, updated, expiry)
+        VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
         "#,
     )
     .bind(base64::encode(hash.0))
@@ -180,6 +181,7 @@ pub async fn insert_payment(payment: &PaymentInfo) -> Result<()> {
     .bind(amt_msat.0.map(|msat| msat as i64))
     .bind(created_timestamp as i64)
     .bind(updated_timestamp as i64)
+    .bind(expiry_timestamp.map(|x| x as i64))
     .execute(&mut conn)
     .await?;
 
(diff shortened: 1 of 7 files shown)


klochowicz:
note: I'm not dealing with invoices created before this changeset (they will have no expiry timestamp) - I will create a separate PR for them, as I suspect more discussions arising from it.

rust/src/db.rs
```diff
@@ -338,6 +343,7 @@ mod tests {
             Flow::Inbound,
             crate::lightning::HTLCStatus::Pending,
             MillisatAmount(Some(100000)),
+            Some(get_timestamp() + 1), // 2 secs expiry time
```
luckysori: ❓ Is it not 1 sec?

rust/src/lightning.rs
```diff
@@ -1152,6 +1156,7 @@ pub async fn send_payment(invoice: &Invoice) -> Result<()> {
         Flow::Outbound,
         status,
         MillisatAmount(invoice.amount_milli_satoshis()),
+        Some(invoice.duration_since_epoch().as_secs() + invoice.expiry_time().as_secs()),
```
luckysori, line 1160: A bit annoying that the library doesn't give you the timestamp directly!

rust/src/lightning.rs
```diff
@@ -1001,6 +1004,7 @@ async fn handle_ldk_events(
                     Flow::Inbound,
                     HTLCStatus::Succeeded,
                     MillisatAmount(Some(*amount_msat)),
+                    None,
```
luckysori, line 1008: 🙃 It's weird to have to deal with this here because the payment has already succeeded.

rust/src/db.rs
```diff
@@ -323,6 +324,36 @@ pub async fn load_payments() -> Result<Vec<PaymentInfo>> {
...
+                    }
+                })
+                .is_some()
+        })
+        .cloned()
+        .collect::<Vec<PaymentInfo>>();
+
+    for payment in newly_expired {
+        update_payment(&payment.hash, HTLCStatus::Failed, None).await?;
+    }
```
luckysori, line 353: 🙃 Using `iter_mut().for_each()` might be a bit simpler.

rust/src/db.rs
```diff
@@ -323,6 +324,36 @@ pub async fn load_payments() -> Result<Vec<PaymentInfo>> {
...
+                        None
+                    }
+                })
+                .is_some()
+        })
+        .cloned()
+        .collect::<Vec<PaymentInfo>>();
+
+    for payment in newly_expired {
+        update_payment(&payment.hash, HTLCStatus::Failed, None).await?;
```
luckysori: 🙃 It could be nice to differentiate between failed payments and expired payments in the UI.

rust/src/db.rs
```diff
@@ -353,9 +384,11 @@ mod tests {
             .set_default()
     }
 
-    async fn ensure_init_fresh_db() -> Result<()> {
+    // Prepare DB for the test. Each test should have DB with unique name to
+    // avoid race conditions.
```
luckysori, line 388: 🙃 What about a random name? It's harder to make a mistake then and easier to use.

rust/src/lightning.rs
```diff
@@ -200,6 +200,7 @@ pub enum HTLCStatus {
     Pending,
     Succeeded,
     Failed,
+    Expired,
```
luckysori, line 203: 👍 Nice!

luckysori (approved):
Nice job! I don't think any of my comments are critical.

rust/src/db.rs
luckysori: This changes in a latter commit anyway.

rust/src/lightning.rs
klochowicz: there's definitely potential for better types and better API here.
I believe the awkwardness is that you might receive a payment without creating an invoice beforehand, hence you only learn about it when you get it.

rust/src/db.rs
klochowicz: I don't mind re-working it, I believe it's exactly the same under-the-hood for this syntactic sugar :)

rust/src/db.rs
klochowicz: if there were more than 2 tests, then yes :P didn't want to invest more time than absolutely needed at this point ;D

holzeis (approved):
LGTM 👍

rust/src/db.rs
klochowicz: I have added capability for this in my changeset :-)- you can diferrentiate it the payment details (a new icon can be added with a one-liner in Dart after this gets in, I just didn't come up with any)

rust/src/db.rs
klochowicz: I don't think I can use `for_each` here, as async closures are still unstable....

luckysori: Yep, I didn't notice the `await` below! Definitely better to go for a for loop here.

PR #532, opened 2022-12-05
Version and network display in settings and logs

### LINKED ISSUE #521: Display the version the user is running in the app, and add it to metadata when sending feedback
Without the version the user is running it can be very hard to trace problems.
Given that we are not sure which version exactly will make it on the user's phone we should send the version and also tell it to the user.

da-kami:
resolves #521

Screenshot 2022-12-05 at 6 48 18 pm

image

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -10,6 +10,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:provider/provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
+import 'package:ten_ten_one/app_info_change_notifier.dart';
 import 'package:ten_ten_one/bridge_generated/bridge_definitions.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_offer_change_notifier.dart';
 import 'package:ten_ten_one/cfd_trading/cfd_order_confirmation.dart';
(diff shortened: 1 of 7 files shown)


lib/settings.dart
```diff
@@ -66,12 +74,17 @@ class _SettingsState extends State<Settings> {
     return TableRow(
```
holzeis, line 74: ❓ I know its not your changes, but why are we using table row here instead of our tto_table?

lib/settings.dart
```diff
@@ -39,6 +41,10 @@ class _SettingsState extends State<Settings> {
 
   @override
   Widget build(BuildContext context) {
+    final appInfoChangeNotifier = context.read<AppInfoChangeNotifier>();
```
holzeis, line 44: nit: if you only intend to read the app info, there is no need to implement the change notifier, resp. update listeners as this will have no effect here.

holzeis (approved):
I've added minor remarks, nothing critical and no need to action! Approved if you want to go directly for the merge.

da-kami:
@holzeis if you find the time feel free to action the comments, I basically agree with all changes, happy to use `TtoRow` instead as well.
Otherwise I'll fix this up tomorrow morning.

lib/main.dart
```diff
@@ -274,6 +278,14 @@ class _TenTenOneState extends State<TenTenOneApp> {
     try {
       await setupRustLogging();
 
+      PackageInfo packageInfo = await PackageInfo.fromPlatform();
+      String version = packageInfo.version;
+
+      FLog.info(text: "TenTenOne Version:" + version);
```
luckysori, line 284: 🙃 We might want a space here.

```suggestion
      FLog.info(text: "TenTenOne Version: " + version);
```

luckysori (approved):
LGTM!

da-kami:
Screenshot 2022-12-06 at 11 47 46 am

PR #536, opened 2022-12-05
fix: Scoped wallet lock only for clone

holzeis:
That will hopefully free up resources on the maker.


klochowicz (approved):
makes sense, let's give it a go!

PR #540, opened 2022-12-05
feat: Reflect disconnected maker

### LINKED ISSUE #525: `Open channel` card appears on restart
When restarting the app with an _already open channel_, the user is first prompted to open a channel via the usual card and then the spinner appears on the card.

Also, any time we lose connection with the maker, the card with the spinner reappears.

I think we need to adjust the logic so that this doesn't happen, as it can be very confusing.

### LINKED ISSUE #535: Show disconnected maker
Currently the open channel card, returns to the inititalizing state if the maker is disconnected. This is because we currently can't really distinguish between a channel creating and a disconnected channel. Both times the channel exist but its not useable. We even use that state to trigger a reconnect.

By using additionally the short_channel_id (if set) we could distinguish if the channel has been already created or its still in creating. Correspondingly we could show a channel creation and the maker connection separately.

Image

holzeis:
Screenshot 2022-12-05 at 13 07 04

- Uses `is_channel_ready` to distinguish between the establishing and disconnected state. According to the docs is_channel_ready will be true after the first confirmation and if there has been at least once a connection.
- Does not set a default channel state, preventing incorrect initial states upon start up.
- Adapts all warning messages accordingly to reflect the new disconnected channel state.

resolves #535
resolves #525

### DIFF
--- a/lib/wallet/channel_change_notifier.dart
+++ b/lib/wallet/channel_change_notifier.dart
@@ -2,14 +2,86 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:ten_ten_one/bridge_generated/bridge_definitions.dart';
+import 'package:ten_ten_one/cfd_trading/validation_error.dart';
 import 'package:ten_ten_one/ffi.io.dart' if (dart.library.html) 'ffi.web.dart';
+import 'package:ten_ten_one/wallet/action_card.dart';
+import 'package:ten_ten_one/wallet/open_channel.dart';
 
 class ChannelChangeNotifier extends ChangeNotifier {
-  ChannelState state = ChannelState.Unavailable;
+  ChannelState? state;
 
   bool isInitialising() => state == ChannelState.Establishing;
 
+  bool isDisconnected() => state == ChannelState.Disconnected;
+
   bool isAvailable() => state == ChannelState.Available;
+
+  bool isUnavailable() => state == ChannelState.Unavailable;
+
+  Message? validate() {
+    if (isUnavailable()) {
+      return Message(
+          title: 'No channel with 10101 maker',
+          details: 'You need an open channel with the 10101.',
+          type: AlertType.warning);
+    }
+
+    if (isInitialising()) {
+      return Message(
+          title: 'Channel not confirmed',
+          details: 'Please wait until your channel has 1 confirmation.',
+          type: AlertType.warning);
+    }
+
+    if (isDisconnected()) {
+      return Message(
+          title: 'Disconnected maker',
+          details: 'You are currently not connected to the maker.',
+          type: AlertType.warning);
+    }
+
+    return null;
+  }
+
+  CardDetails? buildCardDetails() {
+    switch (state) {
+      case ChannelState.Unavailable:
+        return CardDetails(
+            route: OpenChannel.route,
+            title: "Open Channel",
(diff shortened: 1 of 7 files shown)


lib/wallet/channel_change_notifier.dart
```diff
@@ -2,15 +2,87 @@ import 'dart:async';
...
+  bool isDisconnected() => state == ChannelState.Disconnected;
+
   bool isAvailable() => state == ChannelState.Available;
 
+  bool isUnavailable() => state == ChannelState.Unavailable;
+
+  Message? validate() {
+    if (isUnavailable()) {
+      return Message(
+          title: 'No channel with 10101 maker',
```
da-kami, line 24:
```suggestion
          title: 'No channel with 10101',
```

lib/wallet/channel_change_notifier.dart
```diff
@@ -2,15 +2,87 @@ import 'dart:async';
...
+            title: "Establishing Channel",
+            subtitle:
+                "Waiting for channel to get established. This may take a while, waiting for one confirmation.",
+            disabled: true,
+            icon: _buildSpinningWheel());
+      case ChannelState.Available:
+        return null;
+      case ChannelState.Disconnected:
+        return CardDetails(
+            title: "Disconnected Maker",
```
da-kami, line 65:
```suggestion
            title: "Disconnected from 10101",
```

lib/wallet/channel_change_notifier.dart
```diff
@@ -2,15 +2,87 @@ import 'dart:async';
...
+            subtitle: "Trying to reconnect to the 10101 maker.",
```
da-kami, line 66:
```suggestion
            subtitle: "Trying to reconnect to the 10101.",
```

da-kami (approved):
Definitely an improvement, not sure if all scenarios are covered, but I think it's definitely better UX then what was there before!

lib/wallet/channel_change_notifier.dart
```diff
@@ -2,15 +2,87 @@ import 'dart:async';
...
 
   bool isInitialising() => state == ChannelState.Establishing;
 
+  bool isDisconnected() => state == ChannelState.Disconnected;
+
   bool isAvailable() => state == ChannelState.Available;
 
+  bool isUnavailable() => state == ChannelState.Unavailable;
+
+  Message? validate() {
```
luckysori, line 21: The word `validate` is quite vague in this context.

```suggestion
  Message? statusMessage() {
```

lib/wallet/channel_change_notifier.dart
```diff
@@ -2,15 +2,87 @@ import 'dart:async';
...
+      return Message(
+          title: 'Channel not confirmed',
+          details: 'Please wait until your channel has 1 confirmation.',
+          type: AlertType.warning);
+    }
+
+    if (isDisconnected()) {
+      return Message(
+          title: 'Disconnected maker',
+          details: 'You are currently not connected to the maker.',
```
luckysori, line 39: If we want to avoid the "maker" terminology we might want to say

```suggestion
          title: 'Not connected to 10101',
```

The `details` section feels a bit redundant here.

da-kami:
I'll action my comments and then we can go for merge!

luckysori (approved):
Some tweaks, but this should be much better already.

PR #543, opened 2022-12-06
fix: Persist seed backup state over restarts

klochowicz:
Store the value in flutter shared preferences so that we don't prompt the user
to backup the seed after each restart.


luckysori (approved):
If it's tested let's get it in!

Issue #537, opened 2022-12-05
Incorrect input validation on open channel


holzeis:
![Image](https://user-images.githubusercontent.com/382048/205614706-015544b9-6b30-49a0-887a-505b96319715.gif)

PR #544, opened 2022-12-06
Revert "fix: Set initial value to takerChannelAmount"

da-kami:
Sorry, this is actually incorrect!
The maximum amount should not be set to the max channel amount of `500.000` initially, the maximum changes depending on the balance changing!

Reverts itchysats/10101#538

### DIFF
--- a/lib/wallet/open_channel.dart
+++ b/lib/wallet/open_channel.dart
@@ -24,7 +24,7 @@ class OpenChannel extends StatefulWidget {
 }
 
 class _OpenChannelState extends State<OpenChannel> {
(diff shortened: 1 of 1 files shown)


da-kami:
We have to fix this by setting the `takerChannelAmount` based on the computed max amount, not based on the `OpenChannel.maxChannelAmount`

da-kami:
Closing this and fixing it on `main` :)

da-kami:
Reopening because I don't understand how https://github.com/itchysats/10101/issues/537 was a bug - it looks like the behavior was correct? @holzeis what was the problem in https://github.com/itchysats/10101/issues/537 ?

Issue #530, opened 2022-12-05
Check if the feedback form can be configured to default to `email` for sharing options


da-kami:
Related to https://github.com/itchysats/10101/issues/513

On Android we get a `Share` menu which is not intuitive.

Check if it's possible to configure the feedback form to be more specific to share  the feedback via email.

da-kami:
I found a way to specify a fallback that would at least allow us to send text (and the user could attach additional images). This is obviously not as convenient, but better than nothing.

PR #545, opened 2022-12-06
Add fallback to `mailto` URI if `FlutterEmailSender` fails

da-kami:
We unfortunately cannot attach images then, but the user can still send text feedback.

This helps to some extend, but does not solve https://github.com/itchysats/10101/issues/530 on my phone.
It might help with iOS - maybe you can give it a try @luckysori


holzeis (approved):
LGTM 👍

PR #546, opened 2022-12-06
Set taker channel amount to current max

da-kami:
It's tricky to initialize it in the `init` function because we don't know the balance then. Thus, we set it based on the current maximum value if it is null.

Maybe there is a better solution to this, my brain is not working very well at the moment 😅

### DIFF
--- a/lib/wallet/open_channel.dart
+++ b/lib/wallet/open_channel.dart
@@ -24,7 +24,7 @@ class OpenChannel extends StatefulWidget {
 }
 
 class _OpenChannelState extends State<OpenChannel> {
-  int takerChannelAmount = OpenChannel.maxChannelAmount;
+  late int takerChannelAmount;
   bool submitting = false;
 
   final _formKey = GlobalKey<FormState>();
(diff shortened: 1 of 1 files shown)


da-kami:
Tested it locally with both a balance that is > 500000 and a balance of 1000.

lib/wallet/open_channel.dart
```diff
@@ -24,20 +24,25 @@ class OpenChannel extends StatefulWidget {
...
-    super.dispose();
+    final totalBalance = bitcoinBalance.confirmed.asSats +
+        bitcoinBalance.pendingInternal.asSats +
+        bitcoinBalance.pendingExternal.asSats;
+
+    final maxTakerChannelAmount = (totalBalance).clamp(0, OpenChannel.maxChannelAmount);
+
+    takerChannelAmount = maxTakerChannelAmount;
+
+    super.initState();
```
holzeis, line 45: ❓ shouldn't we call super.initState() at first?

lib/wallet/open_channel.dart
```diff
@@ -24,20 +24,25 @@ class OpenChannel extends StatefulWidget {
...
+    final bitcoinBalance = context.read<BitcoinBalance>();
 
-  @override
-  void dispose() {
-    super.dispose();
+    final totalBalance = bitcoinBalance.confirmed.asSats +
+        bitcoinBalance.pendingInternal.asSats +
+        bitcoinBalance.pendingExternal.asSats;
+
+    final maxTakerChannelAmount = (totalBalance).clamp(0, OpenChannel.maxChannelAmount);
```
holzeis, line 41: nit: I don't think these brackets are needed here.

holzeis (approved):
Left some remarks, but go for the merge if it works for you!

lib/wallet/open_channel.dart
holzeis, line 41: nit: you could directly set the `takerChannelAmount` here, instead of using `maxTakerChannelAmount`

PR #550, opened 2022-12-06
feat: Use stream to push data to flutter

### LINKED ISSUE #427: Bitcoin balance and transactions are not synchronised
A small UI inconsistency:

https://user-images.githubusercontent.com/9418575/204429513-af9ed3c6-470c-4c0b-af96-d2048baf5888.mp4

The transaction appears before the balance is updated.

### LINKED ISSUE #524: Hanging on splash screen for too long
It can take a while for the splash screen to go away. In the meantime, it is hard to know if the app is actually working correctly or just broken.

I would add a spinner or something similar.

holzeis:
Uses a stream to push events to the flutter UI instead of regularily polling the rust api. This should also safe some resources, as no thread will be used for every polling request. One thread for our run loop.

**Event Types**

 - **Init**: Sends init messages to the flutter app, reporting about the initialization progress. resolves #524 
 - **Ready**: Indicates the app is ready to use, after starting the ldk node, fetching an offer and balance
 - **Offer**: Pushes an offer to the flutter ui
 - **WalletInfo**: Sends a holistic wallet info update to the flutter UI. This will also fix the issue that the balance was sometimes updated after the history and vice versa. resolves #427 

stream


rust/src/api.rs
```diff
@@ -139,13 +115,121 @@ pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
...
+                        }
+                        Err(e) => {
+                            tracing::error!(?e, "Failed to build wallet info");
+                        }
+                    }
+                    wallet_info_counter = 0;
+                }
+
+                wallet_sync_counter += 5;
+                wallet_info_counter += 5;
```
luckysori, line 229: 🤔 Depending on what actions were performed in the current loop, a lot more than 5 seconds may have passed.

In general, it's unfortunate that we cannot have separate loops for each task, as the logic would be simpler and less error-prone.

rust/src/offer.rs
```diff
@@ -13,29 +10,32 @@ pub struct Offer {
     pub index: f64,
 }
 
-pub async fn get_offer() -> Result<Option<Offer>> {
+pub async fn get_offer() -> Option<Offer> {
     let client = reqwest::Client::builder()
         .timeout(crate::config::TCP_TIMEOUT)
-        .build()?;
+        .build()
+        .expect("reqwest client to build");
```
luckysori, line 17: ❓ So this will never panic?

rust/src/offer.rs
```diff
@@ -13,29 +10,32 @@ pub struct Offer {
...
+            return None;
         }
     };
 
     if response.status() == StatusCode::NOT_FOUND
         || response.status() == StatusCode::INTERNAL_SERVER_ERROR
     {
-        let response = response.text().await?;
-        bail!("Failed to fetch offer: {response}")
+        tracing::warn!("Failed to fetch offer");
```
luckysori, line 30: 🔧 The response context might still be useful for the log.

lib/main.dart
```diff
@@ -104,17 +104,33 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
+              body: Container(
+                  color: Colors.white,
+                  child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
+                    Center(child: Image.asset('assets/10101.finance_logo_1500x1500.png')),
+                    Center(
+                        child: Text(
+                      message,
+                      style: const TextStyle(fontSize: 18),
+                    )),
+                  ]))));
```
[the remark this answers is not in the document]
holzeis: well, yes and no. This peace was to display the init messages during the in app splash screen. However, having this here resulted in a white screen - that's why it was changed again here - c65d7577e63bfd1fcf0b6ee11d9fa0601c20a577

I could clean up the commits a bit if you like.

rust/src/api.rs
```diff
@@ -139,13 +115,121 @@ pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
...
+}
+
+#[tokio::main(flavor = "current_thread")]
+pub async fn run(stream: StreamSink<Event>, app_dir: String) -> Result<()> {
+    let network = config::network();
+    anyhow::ensure!(!app_dir.is_empty(), "app_dir must not be empty");
+    stream.add(Event::Init(format!("Initialising {network} wallet")));
+    wallet::init_wallet(Path::new(app_dir.as_str()))?;
+
+    stream.add(Event::Init("Initialising database".to_string()));
```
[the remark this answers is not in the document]
holzeis: Yes, these messages represent the init messages shown during the splash screen until the app is ready.

https://github.com/itchysats/10101/blob/c65d7577e63bfd1fcf0b6ee11d9fa0601c20a577/lib/main.dart#L317-L318

https://github.com/itchysats/10101/blob/c65d7577e63bfd1fcf0b6ee11d9fa0601c20a577/lib/main.dart#L144-L154

rust/src/api.rs
holzeis: that's very true. For the get offer and get balance, this delay is actually not noticeable. However, the wallet sync could become problematic (as it seems to take more time the more transactions are performed). We could time the sync and remove it from the timeout? (very hacky) or invest into a dedicated thread for syncing (note, i am still not completely sure how many threads are available to use - the limitation might just apply to bridge calls and not to threads spawned from within a rust function - i have to check)

rust/src/offer.rs
holzeis: I guess? looks odd to me that a builder should fail, but I am not sure.

rust/src/api.rs
holzeis: just had a chat with @klochowicz about the topic. The 4 threads are only limited to the bridge calls and pose no limits on the rust code in general. So we can spawn dedicated threads for syncing.

rust/src/offer.rs
holzeis: https://github.com/itchysats/10101/pull/550/commits/a0ea0da977c31ccd59c1a12da7f58b684100b1bc

rust/src/offer.rs
holzeis: https://github.com/itchysats/10101/pull/550/commits/a0ea0da977c31ccd59c1a12da7f58b684100b1bc

rust/src/api.rs
holzeis: https://github.com/itchysats/10101/pull/550/commits/cc7dccd2fdcfb3216744ddcfcd10ce3dbdaffa36

lib/main.dart
```diff
@@ -289,35 +298,31 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
+        if (event is Event_Ready) {
+          tentenoneChangeNotifier.ready();
+        } else if (event is Event_Offer) {
+          cfdOffersChangeNotifier.update(event.field0);
+        } else if (event is Event_WalletInfo) {
+          walletChangeNotifier.update(event.field0);
+        } else if (event is Event_Init) {
+          tentenoneChangeNotifier.set(event.field0);
+        }
       });
```
klochowicz, line 320: A switch statement would be better, with logging an error for unhandled event. (or crashing the whole thing :P I for one would be for explicit handling of all of these.)

lib/models/wallet_info_change_notifier.dart
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  List<PaymentHistoryItem> bitcoinHistory() {
+    return history.where((element) => element.type.isBitcoin()).toList();
+  }
+
+  List<PaymentHistoryItem> lightningHistory() {
+    return history.where((element) => element.type.isLightning()).toList();
+  }
+}
```
klochowicz, line 103: are you not planning to expose total `history` ? (if this is the case, there's little point in creating a single list with them, only to split them up again...)

rust/src/api.rs
```diff
@@ -139,29 +117,132 @@ pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
...
+    WalletInfo::build_wallet_info().await
+}
+
 #[tokio::main(flavor = "current_thread")]
-pub async fn run_ldk() -> Result<()> {
+pub async fn run(stream: StreamSink<Event>, app_dir: String) -> Result<()> {
+    let network = config::network();
+    anyhow::ensure!(!app_dir.is_empty(), "app_dir must not be empty");
+    stream.add(Event::Init(format!("Initialising {network} wallet")));
+    wallet::init_wallet(Path::new(app_dir.as_str()))?;
```
klochowicz, line 176: I'm afraid that with one single `run()` method we might have trouble controlling the invariants established  in #558 
(not impossible, but it's hidden deeper)

rust/src/api.rs
```diff
@@ -139,13 +115,121 @@ pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
...
+            .join(network.to_string())
+            .join("taker.sqlite"),
+    )
+    .await?;
+
     tracing::debug!("Starting ldk node");
+    stream.add(Event::Init("Starting full lightning node".to_string()));
     match wallet::run_ldk().await {
         Ok(_background_processor) => {
+            stream.add(Event::Init("Fetching an offer!".to_string()));
```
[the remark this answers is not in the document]
klochowicz: will this not result in sending `Event` on every offer though?

rust/src/api.rs
klochowicz: I would like to note that the commit mentioned above spawns dedicated tasks on the same single-threaded runtime - not separate threads.

klochowicz (commented):
There's relatively small amount of the code inside this PR that directly contributes to sending initial data on the stream (to convey init message state).

I feel like it would have been heaps easier to review if it was partitioned into a refactor of naming / placement of things, if the intention was to merge this.

holzeis:
> I feel like it would have been heaps easier to review if it was partitioned into a refactor of naming / placement of things, if the intention was to merge this.

haha definitely, I got a little bit carried away with that PR. Happy to close that one and create smaller follow up PRs 🙂

lib/main.dart
holzeis: I tried with a switch, but did not find a way of switching over an instance of. Do you have a suggestion?

lib/models/wallet_info_change_notifier.dart
holzeis: I just move the code as is into a single change notifier, since this data was now coming from one model `WalletInfo`.

rust/src/api.rs
holzeis: can you elaborate on that?

lib/main.dart
klochowicz: aaaaaa. well,  in that case, mabye we should just log an error on `else`? especially if we can print the instance of an unrecognised event?

rust/src/api.rs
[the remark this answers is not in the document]
holzeis: no, only once after the ldk node has started. Afterwards the we will end up in the endless loop - see below.

rust/src/api.rs
holzeis: I guess that's due to this peace 

```
```

Can we use another flavor?

lib/main.dart
holzeis: sounds good!

rust/src/api.rs
klochowicz: I've never researched how a different flavour would work with `frb`.

holzeis:
as discussed with @klochowicz I'll close this PR in favor of multiple smaller ones.

- https://github.com/itchysats/10101/pull/566
- https://github.com/itchysats/10101/pull/567
- https://github.com/itchysats/10101/pull/568
- https://github.com/itchysats/10101/pull/565

PR #552, opened 2022-12-07
Display `confirmed` Bitcoin balance instead of `total`

da-kami:
Displaying the total balance was a great idea, but it causes awkwardness because we then prominently display the maker amount being sent to the taker in the balance upon opening a channel. We cannot easily overcome this problem in the current version, so it's better to display only the confirmed balance prominently.

Additionally this causes awkwardness when we run into a channel force close, because the balance is not `0`, but when trying to open a new channel we can only use `0`, because our displayed balance is actually not usable.

---

@luckysori great composition, I think this is literally the only place I have to change :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -652,6 +652,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dart-sys"
+version = "2.0.1"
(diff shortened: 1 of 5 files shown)


luckysori (approved):
LGTM, this is the right choice!

PR #555, opened 2022-12-07
Use async mutexes

klochowicz:
With sync mutex, if a lock was held at the time another thread requested access
to the wallet, the request would return an error. This resulted in making it the
caller's responsibility to reschedule the request.

With an async mutex, the call will wait until the previous lock goes
out-of-scope and then acquire the lock.

This is not a problem, because we only acquire the lock to share a access to
underlying resources by calling `clone()`, therefore there should be no deadlocks.

The small price we pay for it is the overhead of `current_thread` tokio runtimes
in a few more places in the Rust API for flutter.

### DIFF
--- a/macos/Podfile.lock
+++ b/macos/Podfile.lock
@@ -4,6 +4,8 @@ PODS:
   - FlutterMacOS (1.0.0)
   - mobile_scanner (0.0.1):
     - FlutterMacOS
+  - package_info_plus (0.0.1):
+    - FlutterMacOS
   - path_provider_macos (0.0.1):
     - FlutterMacOS
   - shared_preferences_macos (0.0.1):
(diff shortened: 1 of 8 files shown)


klochowicz:
this solves the "cannot acquire wallet lock" issue

luckysori (commented):
> With sync mutex, if a lock was held at the time another thread requested access
to the wallet, the request would return an error.

I don't think that's how it works. According to the docs of `std::sync::Mutex`:

> This mutex will block threads waiting for the lock to become available.

To me this means that the second thread will just block until the lock is released. Furthermore:

> For a mutex, this means that the [lock](https://doc.rust-lang.org/nightly/std/sync/struct.Mutex.html#method.lock) and [try_lock](https://doc.rust-lang.org/nightly/std/sync/struct.Mutex.html#method.try_lock) methods return a [Result](https://doc.rust-lang.org/nightly/std/result/enum.Result.html) which indicates whether a mutex has been poisoned or not.

Basically, you only get an error if the lock was poisoned. I think we haven't been using this type correctly in this codebase. We shouldn't treat the error returned from `lock` as something that we can easily recover from. The docs even mention that `unwrap`ping is the most common way to handle these.

Maybe we actually shouldn't convert the wallet mutex to a `tokio::sync::Mutex`. The docs say:

> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/mutex/struct.Mutex.html) from the standard library in asynchronous code.
>
> The feature that the async mutex offers over the blocking mutex is the ability to keep it locked across an .await point. This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used. The primary use case for the async mutex is to provide shared mutable access to IO resources such as a database connection. If the value behind the mutex is just data, it’s usually appropriate to use a blocking mutex such as the one in the standard library

klochowicz:
> the docs even mention that unwrapping is the most common way to handle these.

this works really well for synchronous code, but `tokio` catches the panics and the program continues sans the troublesome task. Same with `frb`, this would just end up in a FFI exception in flutter that does nothing except being in the logs. 
I wonder whether there is a way of ensuring reinit of the wallet?

klochowicz:
@luckysori I have gone a bit further and realised that the whole thing with mutexes in `bdk-ldk` was pointless as well. https://github.com/itchysats/bdk-ldk/pull/9

PR #557, opened 2022-12-07
Add back-to-dashboard button to the balance app-bar

### LINKED ISSUE #412: Improve the navigation UX from Trade to Wallet
Currently it's very cumbersome to navigate back from the Trade "app" to the wallet "app". The drawer menu is not intuitive and the user is easily lost.

A potential solution should be evaluated in a **timeboxed** manner.

da-kami:
resolves #412 

If we are on the dashboard the button is not shown. If we are within lightning or bitcoin wallet the button displays home icon. If we are in any other screen that contains the balance on the top (trade, placeholder) we display a `wallet` icon.

This makes it easy to jump back to the dashboard.

### DIFF
--- a/lib/app_bar_with_balance.dart
+++ b/lib/app_bar_with_balance.dart
@@ -1,7 +1,10 @@
 import 'package:flutter/material.dart' hide Divider;
+import 'package:go_router/go_router.dart';
+import 'package:provider/provider.dart';
 import 'package:ten_ten_one/utilities/divider.dart';
 
 import 'package:ten_ten_one/balance.dart';
+import 'package:ten_ten_one/wallet/wallet_change_notifier.dart';
 
 extension BalanceSelectorExtension on BalanceSelector {
   static const preferredHeightVals = {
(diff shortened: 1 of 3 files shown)


da-kami:
Note: We could opt for *always* displaying the `home` icon. I thought about always using the `wallet` icon initially, but that felt weird in the bitcoin/lightning wallet, that's why I came up with the current solution

luckysori (approved):
My brain thanks you!

I have a couple of thoughts:

- I wonder what the standard way of doing this is. Somehow I don't think the back-to-homepage icon goes there. But it's not wrong.
- I'm not sure about having a different icon depending on the context. After all, every version of the icon takes you to the homepage. Maybe it should always be a house.

da-kami:
> * Maybe it should always be a house.

I had the same thought - but when I talked with @klochowicz we had "back to wallet"  in mind - where the "wallet is your home". It's a bit tricky. But now that I look at it I feel having `home` everywhere would be less confusing...

da-kami:
> * I wonder what the standard way of doing this is. Somehow I don't think the back-to-homepage icon goes there. But it's not wrong.

We are for sure not following material design here. The `action` button is supposed to be used for the "most important action the user can do on the screen", [see docs](https://developer.android.com/develop/ui/views/components/appbar/actions). 
But I did not find a better way to deal with this, and given the we already went quite wild with turning our app-bar into the balance it does not feel wrong 😬

klochowicz:
> > * Maybe it should always be a house.
> 
> When I initially talked with @klochowicz we had "back to wallet" in mind - where the "wallet is your home". It's a bit tricky. But now that I look at it I feel having `home` everywhere would be less confusing...

I would still try with a 'wallet' icon - this reinforces the fact that this is wallet balance, and you're going back to the wallet.

Or... actually don't call it the 'dashboard', but 'home'? :)

PR #558, opened 2022-12-07
Do not wrap `Wallet` in mutex

luckysori:
We are never actually mutating the `Wallet` after it has been initialised, so it seems like we don't need a mutex to control access to this resource.

This makes the code a bit simpler and maybe faster since we won't be stuck waiting for some other function to release a mutex guard before we can use the `Wallet`.

---

I've tested it locally a few times and I didn't encounter any trouble.

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -36,7 +36,7 @@ async fn main() -> Result<()> {
             .await
             .expect("lightning node to run");
 
-        let node_info = wallet::get_node_info().expect("To get node id for maker");
+        let node_info = wallet::get_node_info();
         let public_key = node_info.node_id;
         let listening_address = format!("{public_key}@{lightning_p2p_address}");
         tracing::info!(listening_address, "Listening on");
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -145,11 +145,8 @@ pub fn get_wallet_details() -> Result<Json<WalletDetails>, HttpApiProblem> {
             .detail(format!("Internal wallet error: {e:#}"))
     })?;
 
-    let node_info = wallet::get_node_info().map_err(|e| {
-        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
-            .title("Failed get new address")
-            .detail(format!("Internal wallet error: {e:#}"))
-    })?;
+    let node_info = wallet::get_node_info();
+
     Ok(Json(WalletDetails {
         address,
         balance,
@@ -269,16 +266,9 @@ pub struct ChannelConfig {
 }
 
 #[rocket::get("/channel/list")]
-pub async fn get_channel_details() -> Result<Json<Vec<ChannelDetail>>, HttpApiProblem> {
+pub async fn get_channel_details() -> Json<Vec<ChannelDetail>> {
     let list = get_channel_manager()
-        .map_err(|e| {
-            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
-                .title("Failed to create lightning invoice")
-                .detail(format!("{e:#}"))
-        })?
-        .list_channels();
-
-    let vec = list
+        .list_channels()
         .iter()
         .map(|cd| ChannelDetail {
             channel_id: hex::encode(cd.channel_id),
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := help
 PROJECTNAME=$(shell basename "$(PWD)")
 # Default bitcoin network to run the project with `make`
-BITCOIN_NETWORK=regtest
+BITCOIN_NETWORK=testnet
```
da-kami, line 4: ❓ Was this committed on purpose?

luckysori: lol oops

klochowicz (commented):
I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead  of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect). 

if you don't mind, I will create a PR that just removes the mutex.

luckysori:
> I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect).

I must admit I didn't notice that there was a `try_get`. Even so, isn't it an invariant of our programs that the `WALLET` static variable must be `set` before anything else happens?

klochowicz:
> > I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect).
> 
> I must admit I didn't notice that there was a `try_get`. Even so, isn't it an invariant of our programs that the `WALLET` static variable must be `set` before anything else happens?

we could do it this way, but we're not there yet. We actually can have the initial sync failing (and have observed that).
Quick look at shows that we're just proceeding to call `sync` on wallet even after it's uninitialised in flutter (right after catching exception 😅)

happy to work on that afterwards, but I feel it's an orthogonal problem to this PR!


This can be done, and might be a better way - it's just a different

klochowicz:
> > I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect).
> 
> I must admit I didn't notice that there was a `try_get`. Even so, isn't it an invariant of our programs that the `WALLET` static variable must be `set` before anything else happens?

I'm happy to treat it as an invariant 👍
The merit behind treating it as an invariant is that there's no way for the user to deal with such an error, it's mis-use of the API.

I would still probably prefer `try_get().expect('init_wallet() was not called`) so that it's more descriptive, but maybe this is a nit ;D

klochowicz (approved):
happy to try it out with `get()` instead of `try_get()`. Quick glance at the taker and the maker shows that this invariant should hold.

luckysori:
> > > I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect).
> 
> > 
> 
> > I must admit I didn't notice that there was a `try_get`. Even so, isn't it an invariant of our programs that the `WALLET` static variable must be `set` before anything else happens?
> 
> 
> 
> I'm happy to treat it as an invariant 👍
> 
> The merit behind treating it as an invariant is that there's no way for the user to deal with such an error, it's mis-use of the API.
> 
> 
> 
> I would still probably prefer `try_get().expect('init_wallet() was not called`) so that it's more descriptive, but maybe this is a nit ;D

Let's do what you suggest! Feel free to push to this PR and merge it. Otherwise I'll do it tomorrow.

klochowicz:
> > > > I am in favour of removing `Mutex` from the wallet, however I'm not sure about using `get()` instead of `try_get()` - this one can panic in a race condition (if the init takes longer than we expect).
> > 
> > 
> > > 
> > 
> > 
> > > I must admit I didn't notice that there was a `try_get`. Even so, isn't it an invariant of our programs that the `WALLET` static variable must be `set` before anything else happens?
> > 
> > 
> > I'm happy to treat it as an invariant 👍
> > The merit behind treating it as an invariant is that there's no way for the user to deal with such an error, it's mis-use of the API.
> > I would still probably prefer `try_get().expect('init_wallet() was not called`) so that it's more descriptive, but maybe this is a nit ;D
> 
> Let's do what you suggest! Feel free to push to this PR and merge it. Otherwise I'll do it tomorrow.

I did that last night, but forgot to push the commit 😂 coming

luckysori (commented):
Nice! Thanks, @klochowicz!

PR #564, opened 2022-12-08
Only log rust in case of android or ios

holzeis:
I primarily work with the simulator and it bugged me to disable this if all the time. Only logging if on android or ios will not polute the logging if working with native.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,8 +1,8 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:f_logs/f_logs.dart';
 import 'package:feedback/feedback.dart';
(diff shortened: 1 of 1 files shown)


klochowicz (commented):
isn't this new trade-off even worse? 
I'm not interested in the Dart logs when I'm connecting my device, I trust the ones from `tracing` more (they're just clutter ;/)

holzeis:
> isn't this new trade-off even worse? I'm not interested in the Dart logs when I'm connecting my device, I trust the ones from `tracing` more (they're just clutter ;/)

How do you connect to your device? I don't see any rust logs (tracing) when working with a simulator.

PR #567, opened 2022-12-08
feat: In app splash screen with progress messages.

holzeis:
An in-app splash screen using the event stream to push progress messages about initialization.

inappsplashscreen

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -75,7 +75,7 @@ void main() {
     ChangeNotifierProvider(create: (context) => QrScanChangeNotifier()),
     ChangeNotifierProvider(create: (context) => WalletChangeNotifier()),
     ChangeNotifierProvider(create: (context) => cfdOffersChangeNotifier),
-    ChangeNotifierProvider(create: (context) => ChannelChangeNotifier().init()),
+    ChangeNotifierProvider(create: (context) => ChannelChangeNotifier()),
     ChangeNotifierProvider(create: (context) => appInfoChangeNotifier),
   ], child: const TenTenOneApp()));
 }
@@ -319,6 +319,8 @@ class _TenTenOneState extends State<TenTenOneApp> {
               Amount(balance.offChain.pendingClose));
           lightningBalance.update(Amount(balance.offChain.available));
           generatePaymentHistory(walletInfo.lightningHistory, walletInfo.bitcoinHistory);
+        } else if (event is Event_ChannelState) {
+          context.read<ChannelChangeNotifier>().update(event.field0);
         } else {
           FLog.warning(text: "Received unexpected event: " + event.toString());
         }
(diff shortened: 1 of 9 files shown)


lib/main.dart
```diff
@@ -144,6 +142,25 @@ class _TenTenOneState extends State<TenTenOneApp> {
         GoRoute(
             path: '/',
             builder: (BuildContext context, GoRouterState state) {
+              final tentenoneChangeNotifier = context.watch<TenTenOneChangeNotifier>();
+              if (!tentenoneChangeNotifier.isReady()) {
+                Timer(const Duration(milliseconds: 750), () {
+                  // delay removing the splash screen as otherwise the screen will jump due to loading the png
+                  FlutterNativeSplash.remove();
+                });
```
luckysori, line 150: Damn, the transition is almost perfect, but somehow the size of the replacement splash screen is slightly different.

lib/tentenone_change_notifier.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:flutter/material.dart';
+
+class TenTenOneChangeNotifier extends ChangeNotifier {
```
luckysori, line 3: 🙃 Something like that would be a bit more descriptive.

```suggestion
class StartupMessageNotifier extends ChangeNotifier {
```

lib/main.dart
```diff
@@ -321,14 +337,16 @@ class _TenTenOneState extends State<TenTenOneApp> {
...
+          tentenoneChangeNotifier.set(event.field0);
         } else {
           FLog.warning(text: "Received unexpected event: " + event.toString());
         }
       });
     } on FfiException catch (error) {
       FLog.error(text: "Failed to initialise: Error: " + error.message, exception: error);
     } catch (error) {
-      FLog.error(text: "Failed to initialise: Unknown error");
+      FLog.error(text: "Failed to initialise: Unknown error", exception: error);
```
luckysori, line 349: 🔧 Saying that the error is unknown and then showing an exception is confusing.

rust/src/api.rs
```diff
@@ -196,14 +180,34 @@ impl WalletInfo {
 }
 
 #[tokio::main(flavor = "current_thread")]
-pub async fn run_ldk(stream: StreamSink<Event>) -> Result<()> {
```
luckysori, line 199: 🙃 This unification may have been easier to review in a separate commit.

luckysori (approved):
Really cool! It's so much easier to review when decoupled from the other changes.

lib/main.dart
holzeis: yeah, I don't know why though, I am using the exact same image as for the splash screen.

lib/main.dart
holzeis: I guess `Unknown error` because its not an FfiException. This is a catch for any other error. Would you prefer removing the `: Unknown error` part from the message?

luckysori: > Would you prefer removing the : Unknown error part from the message?

Yep, I think it would be slightly clearer.

holzeis:
cherry picked into https://github.com/itchysats/10101/pull/566

PR #568, opened 2022-12-08
chore: refactor balance and history change notifier

holzeis:
Uses a single change notifier instead of three individually. That suites better the single wallet info event update and synchronizes the ui updates.


rust/src/api.rs
```diff
@@ -195,10 +195,19 @@ impl WalletInfo {
...
+    WalletInfo::build_wallet_info().await
+}
+
 #[tokio::main(flavor = "current_thread")]
 pub async fn run_ldk(stream: StreamSink<Event>) -> Result<()> {
     tracing::debug!("Starting ldk node");
     let background_processor = wallet::run_ldk().await?;
+    stream.add(Event::WalletInfo(
+        WalletInfo::build_wallet_info().await.ok(),
+    ));
```
luckysori, line 210: 🔧 This already appears on line 212.

lib/wallet/wallet_change_notifier.dart
```diff
@@ -1,11 +1,119 @@
...
+  set(int index) {
     selectedIndex = index;
     super.notifyListeners();
   }
+
+  Future<void> update(WalletInfo? walletInfo) async {
+    if (walletInfo == null) {
+      // skip empty wallet info update.
+      return;
+    }
```
luckysori, line 30: ❓Why do we need to handle this? Why do we ever emit an empty event?

luckysori (approved):

rust/src/api.rs
holzeis: This is because I want to have the initial data synchronized once, before marking the app ready.

lib/wallet/wallet_change_notifier.dart
holzeis: because I am the `build_wallet_info` can fail and I am transforming that Result into an `Option`. I kind of liked that approach, as it really doesn't matter if now and then a `build_wallet_info` fails. However, currently the error would not get logged, which is a disadvantage.

rust/src/api.rs
luckysori: But you are emitting the `WalletInfo` event _twice_ before making the app ready. I think that's unintentional.

lib/wallet/wallet_change_notifier.dart
luckysori: I do think we should log the error. I'm not sure exactly what component should be responsible for logging it though.

It might be useful to emit the error to the event stream so that the UI can display it. I do agree that we might want to allow some errors to occur before reporting anything to the user, but the best strategy isn't obvious to me.

rust/src/api.rs
holzeis: you are right 👍 thanks!

holzeis:
cherry picked into https://github.com/itchysats/10101/pull/566

Issue #541, opened 2022-12-06
Cannot start latest container


bonomat:
When starting the latest docker image from `main` we get errors: 

```
/usr/bin/maker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /usr/bin/maker)
/usr/bin/maker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /usr/bin/maker)
/usr/bin/maker: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /usr/bin/maker)
```

Tested with `ghcr.io/itchysats/10101:main@sha256:1de84d29ea6f0d3e457f821c261c85534256e849d368c76a5b882ec135400277 `

bonomat:
https://github.com/itchysats/10101/pull/542 did not solve the issue .

bonomat:
I was able to pin it down to the OS on Github being updated. 

For our latest release 0.3.0 we used [ubuntu 20.04.5](https://github.com/itchysats/10101/actions/runs/3600683800/jobs/6065609525) which was now bumped to ubuntu 22.0. See [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners). 

The weird thing is that if we downgrade the runner we get a build error for the flutter bridge: , e.g. see [here](https://github.com/bonomat/10101/actions/runs/3626679417/jobs/6115887916).

klochowicz:
I think it's because the codegen got installed on `ubuntu:latest`, and we need to use `--force` once to override it. or, even simpler solution, just manually delete the cache for that job. I'll try the latter and to re-run the job.

holzeis:
@luckysori can we close that issue? judging from our testnet vm, I guess this issue has been resolved?

```
3aadd95ba230   ghcr.io/itchysats/10101:0.3.2   "/usr/bin/maker --da…"   32 hours ago   Up 32 hours   0.0.0.0:9045->9045/tcp, :::9045->9045/tcp, 0.0.0.0:8888->8000/tcp, :::8888->8000/tcp   tentenone_release
0eb517c1c744   ghcr.io/itchysats/10101:0.3.2   "/usr/bin/maker --da…"   32 hours ago   Up 32 hours   0.0.0.0:8889->8000/tcp, :::8889->8000/tcp, 0.0.0.0:9046->9045/tcp, :::9046->9045/tcp   tentenone_main
```

[...]

PR #570, opened 2022-12-09
fix: use image with glibc 2.35

bonomat:
It looks like this is the version which is being used on github ci and our distroless container does not provide it.

Resolves https://github.com/itchysats/10101/issues/541


bonomat:
Tested in my fork here: https://github.com/bonomat/10101/actions/runs/3654242053

Try with
```
docker pull ghcr.io/bonomat/10101:main
```

klochowicz (approved):

PR #571, opened 2022-12-12
chore: Introduce dedicated runtime

holzeis:
Annotating the function with `#[tokio::main]` killed the runtime after the function completes. Thus we had to block the run function indefinitely for the long living threads to keep going.

This change introduces a runtime managed outside of the function scope and thus allows the `run` function to return bringing the following advantages.

 - We don't block a whole frb worker thread just to run the lightning node, sync tasks, background processor, etc.
 - We are using a multi threaded runtime instead of the current thread
   - allowing to actually join the background processor without blocking all other tasks.
   - making better use of multiple cpu cores.
 - We are not creating a new runtime on every async bridge call.

### DIFF
--- a/rust/src/api.rs
+++ b/rust/src/api.rs
@@ -13,7 +13,6 @@ use crate::offer::Offer;
 use crate::wallet;
 use crate::wallet::Balance;
 use crate::wallet::LightningTransaction;
-use anyhow::anyhow;
 use anyhow::Context;
 use anyhow::Result;
 use flutter_rust_bridge::StreamSink;
@@ -22,13 +21,14 @@ use lightning_invoice::Invoice;
 use lightning_invoice::InvoiceDescription;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
+use state::Storage;
 use std::ops::Add;
 use std::path::Path;
 use std::str::FromStr;
 use std::time::SystemTime;
 use time::Duration;
 pub use time::OffsetDateTime;
-use tokio::try_join;
+use tokio::runtime::Runtime;
 
 pub struct Address {
     pub address: String,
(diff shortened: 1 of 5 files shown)


rust/src/api.rs
```diff
@@ -168,91 +168,114 @@ impl WalletInfo {
...
+        // spawn a connection task keeping the connection with the maker alive.
+        runtime.spawn(async move {
+            let peer_info = config::maker_peer_info();
+            loop {
+                let peer_manager = wallet::get_peer_manager();
+                connection::connect(peer_manager, peer_info).await;
+                // add a delay before retrying to connect
+                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+            }
+        });
```
luckysori, line 232: 🙃 We could extract the bodies of the spawned tasks into functions. It might aid readability.

```suggestion
       runtime.spawn(stay_connected_to_maker());
             
// ...
       
fn async stay_connected_to_maker() {
    let peer_info = config::maker_peer_info();
    loop {
        let peer_manager = wallet::get_peer_manager();
        connection::connect(peer_manager, peer_info).await;
        
        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
}       
```

rust/src/api.rs
```diff
@@ -168,91 +168,114 @@ impl WalletInfo {
...
+        stream.add(Event::ChannelState(get_channel_state()));
+
+        stream.add(Event::Init("Ready".to_string()));
+        stream.add(Event::Ready);
+
+        // spawn a connection task keeping the connection with the maker alive.
+        runtime.spawn(async move {
+            let peer_info = config::maker_peer_info();
+            loop {
+                let peer_manager = wallet::get_peer_manager();
```
luckysori, line 227: 🔧 Is this possible? That way we only call it once.

```suggestion
            let peer_info = config::maker_peer_info();
            let peer_manager = wallet::get_peer_manager();
            loop {
```

luckysori (approved):
Much better imo.

rust/src/api.rs
```diff
@@ -168,91 +168,114 @@ impl WalletInfo {
...
+    static RUNTIME: Storage<Runtime> = Storage::new();
 
-#[tokio::main(flavor = "current_thread")]
-pub async fn refresh_wallet_info() -> Result<WalletInfo> {
-    wallet::sync()?;
-    WalletInfo::build_wallet_info().await
+    if RUNTIME.try_get().is_none() {
+        let runtime = Runtime::new()?;
+        RUNTIME.set(runtime);
+    }
```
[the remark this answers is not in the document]
holzeis: I tried, but I wasn't able to get it to work with a `Result` - do you have a suggestion how this could work?

rust/src/api.rs
holzeis: I had it like this before, but I think clippy complained about the peer_manager being moved or something.

klochowicz: you can probably add `.clone()` to it to bypass clippy

klochowicz (approved):
very interesting!
having an actual multi-threaded runtime that doesn't take any of the threads on frb threadpool is exciting - I'd love to give it a go and see how it works.

PR #572, opened 2022-12-12
feat: Restart failed background processor

holzeis:
Adds and handling in case the background processor stops, by simply trying to start the background processor again.

Some more refactoring could be probably done there, but I tried to keep the changes to a minimum.


luckysori:
Not sure, why the PR includes https://github.com/itchysats/10101/pull/572/commits/d895a5b5b4bae91921c4b11e1731ade2fd5551fa. Maybe you need to rebase on https://github.com/itchysats/10101/pull/571?

rust/src/api.rs
```diff
@@ -268,10 +268,15 @@ pub fn run(stream: StreamSink<Event>, app_dir: String) -> Result<()> {
...
 
         runtime.spawn_blocking(move || {
-            // background processor joins on a sync thread, meaning that join here will block a
-            // full thread, which is dis-encouraged to do in async code.
-            if let Err(err) = background_processor.join() {
-                tracing::error!(?err, "Background processor stopped unexpected");
+            let mut background_processor = background_processor;
+            loop {
+                // background processor joins on a sync thread, meaning that join here will block a
+                // full thread, which is dis-encouraged to do in async code.
```
luckysori, line 274: 🔧 It should be "discouraged"!

rust/src/api.rs
```diff
@@ -268,10 +268,15 @@ pub fn run(stream: StreamSink<Event>, app_dir: String) -> Result<()> {
...
+                if let Err(err) = background_processor.join() {
+                    tracing::warn!(?err, "Background processor stopped unexpected");
```
luckysori, line 276: 🔧 It should be "unexpectedly"!

rust/src/lightning.rs
```diff
@@ -510,6 +527,14 @@ pub fn setup(
         IgnoringMessageHandler {},
     ));
 
+    // Step 17: Initialize routing ProbabilisticScorer
+    let scorer_path = format!("{ldk_data_dir}/scorer");
+    let scorer = Arc::new(Mutex::new(disk::read_scorer(
+        Path::new(&scorer_path),
+        Arc::clone(&network_graph),
+        Arc::clone(&logger),
+    )));
```
luckysori, line 536: ❓ Why did we have to move this?

rust/src/lightning.rs
```diff
@@ -583,20 +599,9 @@ pub fn run_ldk(system: &LightningSystem) -> Result<BackgroundProcessor> {
...
-        system.persister.clone(),
-        invoice_payer.clone(),
-        system.chain_monitor.clone(),
-        system.channel_manager.clone(),
-        GossipSync::p2p(system.gossip_sync.clone()),
-        system.peer_manager.clone(),
-        system.logger.clone(),
-        Some(scorer),
-    );
+    let background_processor = system.start_background_processor();
```
luckysori, line 602: 🔧 I'm not a huge fan of calling this here (inside of `run_ldk`) and also on the outside when restarting the background processor.

Actually, this function appears to be doing only 2 things:

- Setting the `INVOICE_PAYER` static variable, something that should only happen once during the lifetime of the app AFAIK.
- Starting the background processor via the `LightningSystem`.

As such, I think we should get rid of the `run_ldk` function and just do those two things separately when necessary.

luckysori (approved):
I think these changes are actually still part of your other PR?

Once that is cleared up, LGTM!

rust/src/lightning.rs
holzeis: to be able to start the background processor from the properties of the lightning system

rust/src/lightning.rs
holzeis: agreed, but 

> Some more refactoring could be probably done there, but I tried to keep the changes to a minimum.

